PR #3, opened 2019-08-17
[NS-1651] Add parameter to rotate EnvironmentLight texture 180 degree…

lyonsno:
…s around Y axis

## Brief Description
Add rotate180 parameter to THREE.EnvironmentLight class.  This parameter rotates the map 180 degrees around the Y (vertical) axis.  Could potentially be expanded with an arbitrary parameter in degrees or radians.  Adding this now so that we can easily match the 180 rotated cubeMaps that three.js generates from the equirectangular maps that we use.

Not sure about the parameter name.
Performance will be impacted because this requires another iteration over all pixels in the map.  This can be rolled into the performance refactor of the entire envMapCreation.js file.

## Pull Request Guidelines

- [ x ] I have added pull requests labels which describe my contribution.
- [ x ] All existing tests passed.
- [ x ] My code has passed the ESLint configuration for this project.
- [ x ] My change requires modifications to the documentation.


jaxry:
Envmap rotation would be very useful. But I think it doesn't make much sense just to limit it to 180 degrees. Shouldn't we just expose a `rotation` parameter where the user provides a rotation in radians, and the environment map is rotated by that amount? It wouldn't be any more work but would be much more generally applicable.

lyonsno:
> Envmap rotation would be very useful. But I think it doesn't make much sense just to limit it to 180 degrees. Shall we just expose a `rotation` parameter where the user provides a rotation in radians, and the environment map is rotated by that amount? It wouldn't be significantly more work to implement but would be much more generally applicable.

Yeah that would be best for sure.  This is the MVP for a blocker on another ticket, but maybe it's too confusing to merge this and then go in and change the param completely in a little bit.

> As for performance, envmap generation is currently the slowest part of the initialization time by far. Is it a good idea to be slowing down initialization by another several hundred ms instead of focusing on optimizing the envmap file first? I think that should be our highest priority since slow loading times really affect the user experience in my opinion.

I agree about this, again see above for my (admittedly biased) motivation, was thinking that this whole file is gonna get a re-write within a week or so anyway.  I can see how this kind of thing is potentially less justifieable in a public repo though, I'm not sure exactly what the etiquette is there. 

> Edit: If we need to rotate envmaps for Machete, would the cleanest solution be for now to rotate the envmap in Machete before creating the EnvironmentLight? That way we can implement this feature correct and proper (without performance loss) the first time around.
> 

Good point, this is something that occurred to me after pushing. The thing is that THREE.js is really the place where this functionality should live and it keeps juuuust barely not making it into the last several revs.  

> > easily match the 180 rotated cubeMaps that three.js generates from the equirectangular maps that we use.
> 
> Is our environment light backwards by default? If that's the case, why don't we correct it so that both three.js and the environment light orient the envmaps the same way?

It's possible that this is a situation where there isn't a standardized orientation.  I wasn't able to find one.  It does occur to me that if that is the case we should just conform to three.js.

After typing all this out I think I know what I'm gonna do.  As I realized right here^ we should conform to three.js, so I think that would be a good placeholder that would unblock me until rotation is calculated to a specific angle.

Great thoughts as usually @jaxry, thanks :)

jaxry:
I just checked, and [EquirectangularToCubemap](https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/EquirectangularToCubeGenerator.js) also produces an envmap backwards from EnvironmentLight. Since this is the case, I suggest we rotate the EnvironmentLight's envmap by 180 degrees by default. In this case, do we even need this temporary workaround?

jaxry:
What is the status of this PR?

jaxry:
Closing the PR for now since there might not be much use to rotating an envmap by only 180 degrees.

PR #4, opened 2019-08-20
reverse envmap x y direction

jaxry:
## Brief Description
This PR rotates the orientation of the Environment Light by 180 degrees. Since Environment Light is a new type of light, there's no inherently correct orientation. However, the  [EquirectangularToCubemap](https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/EquirectangularToCubeGenerator.js) class provided by Three.js generates cubemaps that are arbitrarily opposite of the orientation of the Environment Light.

This PR rotates the Environment Light so that equirectangular maps provided to both the WebGLRenderer (by using EquirectangularToCubeMap) and RayTracingRenderer are the same orientation. This should address the issue brought up in https://github.com/hoverinc/ray-tracing-renderer/pull/3.

In addition, I've cleaned up the coordinate conversion code by moving it into its own function.

TODO or DELETE

## Pull Request Guidelines

### DIFF
--- a/src/renderer/glsl/chunks/envmap.glsl
+++ b/src/renderer/glsl/chunks/envmap.glsl
@@ -7,6 +7,12 @@ export default function(params) {
 uniform sampler2D envmap;
 uniform sampler2D envmapDistribution;
 
+vec2 cartesianToEquirect(vec3 d) {
+  float phi = mod(atan(-d.z, -d.x), TWOPI);
+  float theta = acos(d.y);
+  return vec2(phi * 0.5 * INVPI, theta * INVPI);
+}
+
 float getEnvmapV(float u, out int vOffset, out float pdf) {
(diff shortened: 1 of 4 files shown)


src/renderer/glsl/chunks/envmap.glsl
```diff
@@ -7,6 +7,12 @@ export default function(params) {
 uniform sampler2D envmap;
 uniform sampler2D envmapDistribution;
 
+vec2 cartesianToEquirect(vec3 d) {
```
elfrank: would it make sense to rename `d` to something more descriptive?

jaxry: `d` can be any normalized vector pointing to a point on the sphere. I agree the name is bad but I wonder what else we can call it.

PR #5, opened 2019-08-21
Automatically define GLSL constants

jaxry:
## Brief Description
This PR adds a helper function which automatically places `#define` directives in the GLSL code. This gets rid of some messy copy-pasted code that made adding newly defined constants difficult.

## Pull Request Guidelines

### DIFF
--- a/src/renderer/glslUtil.js
+++ b/src/renderer/glslUtil.js
@@ -8,3 +8,15 @@ export function unrollLoop(indexName, start, limit, step, code) {
 
   return unrolled;
 }
+
+export function addDefines(params) {
+  let defines = '';
+
+  for (let [name, value] of Object.entries(params)) {
+    if (value) {
+      defines += `#define ${name} ${value}\n`;
+    }
+  }
+
+  return defines;
+}
--- a/src/renderer/glslUtil.js
+++ b/src/renderer/glslUtil.js
@@ -13,6 +13,8 @@ export function addDefines(params) {
   let defines = '';
 
   for (let [name, value] of Object.entries(params)) {
+    // don't define falsy values such as false, 0, and ''.
+    // this adds support for #ifdef on falsy values
     if (value) {
       defines += `#define ${name} ${value}\n`;
     }
--- a/src/renderer/glsl/chunks/intersect.glsl
+++ b/src/renderer/glsl/chunks/intersect.glsl
@@ -1,16 +1,6 @@
 export default function(params) {
   return `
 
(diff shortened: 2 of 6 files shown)


src/renderer/glslUtil.js
```diff
@@ -7,4 +7,18 @@ export function unrollLoop(indexName, start, limit, step, code) {
...
-}
\ No newline at end of file
+}
+
+export function addDefines(params) {
+  let defines = '';
+
+  for (let [name, value] of Object.entries(params)) {
+    // don't define falsy values such as false, 0, and ''.
+    // this adds support for #ifdef on falsy values
```
carlos-lopez-garces: Couldn't `0` be the legitimate value of a numeric parameter? If the macro is used directly to supply a literal value, as opposed to being checked for definition with `#if defined()` (much like `BVH_COLUMNS` is used in `intersect.glsl`), the expansion will yield nothing and there'll be a compile error most likely.

Do we not expect such kind of parameters?

jaxry: You bring up a good point, and I've been thinking about that. It's definitely possible for there to be some outlying case where you want to use a define of 0 for something. But that's the exception rather than the norm and I can't think of an example off the top of my head. 

The current implementation of simply not defining 0 is a lot more useful since we use that functionality several times already. I reasoned that if we ever need to support 0, I can refactor the code again. In the meantime we're left with greater simplicity in the code with something that works for the foreseeable future.

What do you think? Is that reason good enough to go of off? If not, how might you add support for 0 in a simple way while still allowing for the possibility of *not* defining anything that's zero? If you think the current code is too fragile but don't have other ideas, that's fine too and I can do some more brainstorming.

carlos-lopez-garces: I wouldn’t say it’s fragile; I was just thinking that people that are new to the code and that add their own numeric parameters with legitimate 0s might be puzzled by compile error if they aren’t familiar with how `addDefines` works (after all, the parameter list and the definition of `addDefines` are in 2 separate files); I think we’d avoid that altogether if we add a comment to the parameter list (where it’s passed to the shader) telling readers to go check out the definition of `addDefines` for details.

Sounds like you’ve put a good deal of thought into this, and I don’t have a good answer yet, so I wouldn’t let this prevent you from merging.

elfrank: Valid future concern. Good catch @carlos-lopez-garces .

>  I reasoned that if we ever need to support 0, I can can add that feature when we need it. In the meantime we're left with greater simplicity in the code with something that works for the foreseeable future.

sounds good to me

PR #10, opened 2019-09-11
Build 0.1.1

jaxry:
This PR upgrades the renderer from 0.1.0 to 0.1.1. 

## Changelog:

* https://github.com/hoverinc/ray-tracing-renderer/pull/7 Replace white noise with blue noise for a higher quality render.
* https://github.com/hoverinc/ray-tracing-renderer/pull/6 Fix overly dark shadow catcher visible in reflections of other objects.
* https://github.com/hoverinc/ray-tracing-renderer/pull/4 **(Breaking Change)** Rotate orientation of EnvironmentLight by 180 degrees. The orientation now matches Three.js's [EquirectangularToCubeGenerator](https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/EquirectangularToCubeGenerator.js).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "three.js-ray-tracing-renderer",
-  "version": "1.0.0",
+  "version": "0.1.1",
   "description": "",
   "main": "src/RayTracingRenderer.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "three.js-ray-tracing-renderer",
-  "version": "1.0.0",
```
jaxry: Incorrect version. This should be `0.1.0`

PR #13, opened 2019-09-16
Clean-up before denoising

jaxry:
## Brief Description
This PR is my first pass at refactoring and preparing the code for the denoising pipeline. Nothing serious has changed here and I'm just trying to clean things up before a more serious change down the line.

Most notable refactors:

* Rearrange functions so that the exported function is at the top of the file. This style is used by the majority of JS projects, and in my opinion having the exported function at the top makes the file easier to read when first opening it up.

* Split `RayTracingShader.js` into smaller functions. Details in the PR comments.

* Remove `Object.freeze`. I tried the idea of returning frozen module objects, but I don't think it's a good idea. `Object.freeze` is relatively obscure and simply complicates things. If a user tries to add a property to a frozen object, the added property will be silently ignored, making for a lot of possible confusion. Now we just return a simple object as expected.

As I move forward implementing hybrid rendering and denoising, I will be rewriting the top-level pipeline and GL memory management from scratch. The new system will allow for more dynamic scene changes (e.g. allowing lighting changes without reinitalizing the entire renderer), and the purpose of each file and function should be way more clear than it is now. 

The current code was written largely by me as educational experience since I didn't know what I was doing this first time around. But these small refactors should improve the code a little bit while I work on the new pipeline.

## Pull Request Guidelines


src/renderer/glUtil.js
```diff
@@ -147,8 +118,36 @@ export function makeUniformBuffer(gl, program, blockName) {
...
 
-  return Object.freeze({
+  return {
     set,
     bind
-  });
+  };
+}
+
+function getUniformBlockInfo(gl, program, blockIndex) {
```
jaxry: Move this child function into it's own function for simplicity.

src/renderer/rayTracingShader.js
```diff
@@ -39,180 +37,9 @@ export function makeRayTracingShader({
...
-    return {
-      program,
-      uniforms,
-    };
-  }
-
-  const { program, uniforms } = initScene();
+  const { program, uniforms } = makeProgramFromScene({
+    bounces, fullscreenQuad, gl, optionalExtensions, samplingDimensions, scene, textureAllocator
+  });
```
jaxry: This child function is now its own independent function

src/renderer/rayTracingShader.js
```diff
@@ -296,6 +295,38 @@ function textureDimensionsFromArray(count) {
   };
 }
 
+function makeDataTexture(gl, dataArray, channels) {
```
jaxry: New function to reduce duplicated code.

src/renderer/rayTracingShader.js
```diff
@@ -296,6 +295,38 @@ function textureDimensionsFromArray(count) {
...
+  return makeTexture(gl, {
+    data: padArray(dataArray, channels * textureDim.size),
+    minFilter: gl.NEAREST,
+    magFilter: gl.NEAREST,
+    width: textureDim.columns,
+    height: textureDim.rows,
+  });
+}
+
+function makeTextureArray(gl, textures, gammaCorrection = false) {
```
jaxry: New function to reduce duplicated code.

src/renderer/glsl/chunks/envmap.glsl
```diff
@@ -90,8 +94,8 @@ float envmapPdf(vec2 uv) {
 
   uv *= size;
 
-  float partialX = texelFetch(envmapDistribution, ivec2(1.0 + uv.x, uv.y), 0).g;
-  float partialY = texelFetch(envmapDistribution, ivec2(0, uv.y), 0).g;
+  float partialX = texelFetch(envmapDistribution, ivec2(1.0 + uv.x, uv.y), 0).y;
```
jaxry: Use `y` which is just a syntactic variation of `g`. The code style reflects the idea that `y` should refers to the second channel in a data texture, and `g` refers to a channel which represents a green color intensity

src/renderer/glsl/chunks/sampleGlassSpecular.glsl
```diff
@@ -27,8 +27,8 @@ vec3 sampleGlassSpecular(SurfaceInteraction si, int bounce, inout Ray ray, inout
   initRay(ray, si.position + EPS * lightDir, lightDir);
 
   // advance sample index by unused stratified samples
-  const int usedDimensions = 1;
-  sampleIndex += DIMENSIONS_PER_MATERIAL - usedDimensions;
+  const int usedSamples = 1;
+  sampleIndex += SAMPLES_PER_MATERIAL - usedSamples;
```
jaxry: `SAMPLES_PER_MATERIAL` makes more sense than the former.

src/renderer/fullscreenQuad.js
```diff
@@ -18,8 +18,8 @@ export function makeFullscreenQuad(gl) {
     gl.drawArrays(gl.TRIANGLES, 0, 6);
   }
 
-  return Object.freeze({
+  return {
     draw,
```
santiagoroca: Usually, the best way to treat immutables is a pattern that utilizes proxies and getter and setters. Basically you pass the initial data, and a Proxy that just allows you to access that data is created. You can even throw exceptions if someone tries to use data outside of the initial object structure. That pattern works well in these scenarios!

Let me know if you'd like me to expand or jump in a call to work it out together. That's the tiny bit I can add to this project :sweat_smile:

src/renderer/glsl/chunks/intersect.glsl
```diff
@@ -115,7 +115,7 @@ void surfaceInteractionFromIntersection(inout SurfaceInteraction si, Triangle tr
       dpdu *= invmax;
       dpdv *= invmax;
 
-      vec3 n = 2.0 * texture(normalMap, vec3(uv * materials.diffuseNormalMapSize[normalMapIndex].zw, normalMapIndex)).rgb - 1.0;
+      vec3 n = 2.0 * texture(normalMap, vec3(uv * materials.diffuseNormalMapSize[normalMapIndex].zw, normalMapIndex)).xyz - 1.0;
       n.xy *= materials.roughnessMetalnessNormalScale[materialIndex].zw;
```
santiagoroca: I'm not very familiar with GLSL conventions, so, if no relevant, ignore my question: Isn't for a texture more natural to utilize RGB? I know it makes no technical difference but I'm intrigued on the motivation behind the change

src/renderer/texture/readme.txt
```diff
@@ -1,4 +1,4 @@
-noise.js is a 64x64 resolution precomputed blue noise texture.
+noise.js is a 64x64 pixel precomputed blue noise texture.
 It was created by Christoph Peters and downloaded from http://momentsingraphics.de/BlueNoise.html
```
santiagoroca: nit: pixels?

src/renderer/fullscreenQuad.js
jaxry: I'd like to learn more! Do I get what you're talking about? You return `get`ters and `set`ters to properties instead of the properties themselves, or otherwise return functions that control access to the module instead of returning the mutable state directly? If that's so, how do you return the module object itself so that it's not accidentally mutated? That's why I was using `Object.freeze` before, but that doesn't seem like a great solution, since it silently fails.

src/renderer/glsl/chunks/intersect.glsl
jaxry: The motivation relies on the fact that we are using both textures which encode data, and textures which encode color in the shader, so I'd like a simple way to differentiate the two.

The `texture` sampler always returns a vec4, and in some cases that vec4 contains a spatial 3-dimentional vector. It feels natural to me to access a texel containing a spacial vector via `.xyz`. In cases where the sampler returns a vec4 that represents channels with colors, instead of vectors in space, that's when it feels natural to use `.rgb`.

You actually pointed out a case where it's tricky. The channels of that texture aren't supposed to be used as colors, but they need to be "converted" to normalized vectors before they're usable. Right out of the texture they're in the [0, 1] color range and need to be transformed to the [-1, 1] range. I agree it makes sense to use `.rgb` here.

jaxry:
Is this PR okay to merge?

lyonsno:
I’m heading home right now, and I was planning on reviewing it when I got
there. Will it block you a lot if I don’t get my review in for another
couple of hours?

jaxry:
Oh by all means take your time! I was just wondering if anyone else was planning to review it, so I appreciate that you are.

src/renderer/glUtil.js
```diff
@@ -61,43 +60,15 @@ export function getUniforms(gl, program) {
...
-    return uniforms;
-  }
-
-  const uniforms = getUniformInfo();
+  const uniforms = getUniformBlockInfo(gl, program, blockIndex);
 
   const buffer = gl.createBuffer();
   gl.bindBuffer(gl.UNIFORM_BUFFER, buffer);
-  gl.bufferData(gl.UNIFORM_BUFFER, blockSize, gl.DYNAMIC_DRAW);
+  gl.bufferData(gl.UNIFORM_BUFFER, blockSize, gl.STATIC_DRAW);
```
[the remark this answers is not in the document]
jaxry: Since we're creating and updating the uniform buffer only once, `STATIC_DRAW` describes our use case. It's supposed to let the driver optimize shader use, but in reality I didn't measure a performance difference. Either way, it's the correct value now.

PR #16, opened 2019-09-23
Expand .gitignore

elfrank:
## Brief Description
In this PR, we add additional files to ignore on the Github repository

* `.DS_Store`
* `demos/` folder: this way, we can easily experiment locally with scenes
* `test-scene/` folder: this way, we can keep a local copy of the test
scene.

## Pull Request Guidelines

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,7 @@
+.DS_Store
+
 node_modules
+
+# demo/test scenes
+demos/
+test-scene/


santiagoroca:
Just an idea:

Would it be useful to have a single "local" folder at the root with all of this kind of files that are only meant to be kept locally? For example:

local/test-scene
local/demos
local/tmp/orders

With only ignoring "local/*" we would ignore this and future local files.

jaxry:
Interesting idea Santiago. I agree there's something we can do here to handle these demo files better. I'll think about these options.

It would be nice if we had an automated way of installing all these local files. Putting the files in the repo officially is the easiest thing to do, but the problem with that is that when we `npm install` this repo as a dependency, the dependency will include the megabytes of examples scenes that won't be needed as a dependency.

Do you have any ideas how we can track these files while at the same time excluding them from the renderer when used as a dependency?

santiagoroca:
I did not fully understand the implication of the problem. I wouldn't want to delay a change of a .gitignore for such a small suggestion so feel free to disregard this! 

Two things come to my mind and those might be extreme over-engineer this simple task:

1. Create a separate repo for all these additional files and add them as devDependency. Each time you `npm install` on the repo, it will fetch the newest scenes
2. An alternative might be, instead of using the http-server module to start the application, create a small static-server that also works a proxy, so each time one of these scenes is "fetched" if not present, it'd be downloaded to a tmp folder. Machete uses this same approach for orders.

elfrank:
> Just an idea:
> 
> Would it be useful to have a single "local" folder at the root with all of this kind of files that are only meant to be kept locally? For example:
> 
> local/test-scene
> local/demos
> local/tmp/orders
> 
> With only ignoring "local/*" we would ignore this and future local files.

Yeah, I thought about something like that but the "future" needs are not clear. Keeping it simple for now.

> Interesting idea Santiago. I agree there's something we can do here to handle these demo files better. I'll think about these options.
> 
> It would be nice if we had an automated way of installing all these local files. Putting the files in the repo officially is the easiest thing to do, but the problem with that is that when we `npm install` this repo as a dependency, the install will include the megabytes of examples scenes that aren't needed as a dependency.
> 
> Do you have any ideas how we can track these files while at the same time excluding them from the renderer when used as a dependency?

We could add a `bootstrap.sh` shell script, where we "setup" the repo. We could include fetching the `test-scene`, demos and other relevant files. Something to think about.

PR #17, opened 2019-09-24
Optimize Envmap Creation

lyonsno:
**Current Changes**
Change RGBEToFloat exponentiation to bit shift

Add multiplier to RGBEToFloat to optimizing environment light intensity adjustment

Implement early bail on addLightAtCoordinates once last light texel has been added

Optimize out most 3d angle calculations by implementing a threshold on theta and phi

This is a rough pass so everyone can get a look at what's going on here

**Context**
The ray tracing renderer currently accepts Environment Lights with. HDR maps encoded in RGBE format. It then performs some processing steps on these maps to prepare them for use.  First, the maps are converted from RGBE encoding to Floating point.  Then the values are multiplied by the intensity of the Environment Light.  After that, the lighting from other Three.js lights in the scene are applied on top of the map.

These calculations are performed on the CPU by iterating over every texel in the buffer and modifying its value appropriately.  They must be completed before scene initialization can complete, and must also be run every time lighting values or environment map are changed.

**Description**
These calculations can take a significant amount of time on large maps, and even on smaller 2k maps they delay scene initialization by 1000 ms.  On top of other problems, this makes tweaking lighting values much more difficult.  Moving computation to the GPU has been considered, but memory transfer incurs an overhead of almost 1000 ms as well, so for 2k maps optimization isn't possible that way.  Therefore, we need to optimize these calculations on the CPU so that scenes load more quickly and lighting changes update closer to real-time.

**Goal**
Decrease the time needed to generate the final environment map used by the ray tracing renderer.

## Pull Request Guidelines

    - [x] I have added tests to cover my changes, of which pass.


src/renderer/envMapCreation.js
```diff
@@ -59,36 +59,86 @@ export function addDirectionalLightToEnvMap(light, image) {
...
   const height = image.height;
-
   const xTexels = (floatBuffer.length / (3 * height));
   const yTexels = (floatBuffer.length / (3 * width));
-  // default softness for standard directional lights is 0.95
-  const softness = ("softness" in light && light.softness !== null) ? light.softness : 0.45;
+  const density = 8;
+  // default softness for standard directional lights is 0.01, i.e. a hard shadow
+  const softness = ("softness" in light && light.softness !== null) ? light.softness : 0.01;
+
```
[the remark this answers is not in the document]
lyonsno: good point!

lyonsno:
Ok some performance metrics!!

The optimizations implemented in this PR increase performance more significantly with sharper (softness closer to 0.0) directional lights.  This is because many of the optimizations involve using approximations and early bail out conditions to avoid running the more expensive computations on as many texels as possible. Softer lights by definition cover more area on the map, and so fewer computations can be skipped. Still, the performance increase is at least an order of magnitude faster until softness begins to approach 0.4.  Fortunately for us, we almost certainly won't be using softness values that high. Our current look uses ~0.1.  Anyway, here's an example of shadows at 0.1, 0.4, and 0.8 softness, respectively:

0.1
Screen Shot 2019-09-26 at 2 59 14 PM

0.4
Screen Shot 2019-09-26 at 2 41 10 PM

0.8
Screen Shot 2019-09-26 at 2 57 21 PM


Now for the metrics! This is a good way to see how performance compares across different softness values.  I've only included one profile image for the original version, since it follows the same logic regardless of shadow softness.

Original
old_average
softness=0.0 ~ 1540 ms
softness=0.25 ~ 1260 ms
softness=0.5 ~ 1180 ms
softness=0.75 ~ 1210 ms
softness=1.0 ~ 1600 ms

Optimized (This PR)
softness=0.0 ~ 100 ms
new_00

softness=0.25 ~ 140 ms
new_025

softness=0.5 ~ 170 ms
new_05

softness=0.75 ~ 810 ms
new_075

softness=1.0 ~ 1200 ms
new_1

src/renderer/envMapCreation.js
```diff
@@ -59,36 +59,90 @@ export function addDirectionalLightToEnvMap(light, image) {
...
-      const currentSphericalCoords = equirectangularToSpherical(i, j, width, height);
-      const falloff = getIntensityFromAngleDifferential(originSphericalCoords, currentSphericalCoords, softness);
-      const intensity = light.intensity * falloff;
+      bufferIndex = j * width + i;
+      currentCoords = equirectangularToSpherical(i, j, width, height, currentCoords);
+      falloff = intensityFromAngleFunction(originCoords, currentCoords, softness, threshold);
+
+      if(falloff > 0) {
+        encounteredX = true;
+        encounteredY = true;
```
jaxry: Could `encounteredX` and `encounteredY` act one boolean? If not, is there a way to simplify this logic? I don't exactly follow it.

src/renderer/envMapCreation.js
```diff
@@ -97,16 +151,22 @@ function angleBetweenSphericals(originCoords, currentCoords) {
...
+  // TODO: possibly clean this up and optimize it
+  //
+  // This function was arrived at through experimentation, it provides good 
+  // looking results with percieved softness that scale relatively linearly with
+  //  the softness value in the 0 - 1 range
+  //
+  // For now it doesn't incur too much of a performance penalty because for most of our use cases (lights without too much softness)
+  // the threshold cutoff in getIntensityFromAngleDifferential stops us from running it too many times
 function getFalloffAtAngle(angle, softness) {
   const softnessCoeficient = Math.pow(2, 14.5 * Math.max(0.001, (1.0 - clamp(softness, 0.0, 1.0))));
```
jaxry: We already saw in the case of `rgbeToFloat` how slow `Math.pow` is. Have you measured the performance difference of caching the coefficient?

Also, `coefficient` instead of `coeficient` :grin:

src/renderer/rgbeToFloat.js
```diff
@@ -1,18 +1,23 @@
...
     const g = buffer[4 * i + 1];
     const b = buffer[4 * i + 2];
     const a = buffer[4 * i + 3];
-    const e = 2 ** (a - 128);
-    floatBuffer[3 * i] = r * e / 255;
-    floatBuffer[3 * i + 1] = g * e / 255;
-    floatBuffer[3 * i + 2] = b * e / 255;
+
+    const exponent = a - 128;
+    const e = exponent > 0 ? 1 << exponent : 1 / (1 << (-1 * exponent));
```
jaxry: Could we add a comment that explains this is equivelant to `e = Math.pow(2, exponent)`?

src/renderer/rgbeToFloat.js
```diff
@@ -1,18 +1,23 @@
 // Convert image data from the RGBE format to a 32-bit floating point format
 // See https://www.cg.tuwien.ac.at/research/theses/matkovic/node84.html for a description of the RGBE format
-export function rgbeToFloat(buffer) {
+// Optional multiplier argument for performance optimization
+export function rgbeToFloat(buffer, multiplier = 1) {
   const texels = buffer.length / 4;
   const floatBuffer = new Float32Array(texels * 3);
-
+  const imageCoords = [0, 0];
```
jaxry: Not needed?

src/renderer/envMapCreation.js
lyonsno: so this is weird.  When I first read this comment I was excited because it seemed like a guaranteed speedup.  After profiling though, caching the coef`f`icient  actually seems to incur a performance penalty??? Is that possible?

jaxry: Are you accessing it via a closure or via a direct function argument? It could be that if the variable is defined far enough up the function stack and accessed via closures, the JS engine needs to do some trickery to access the variable in the loop, which is slower than recomputing it each time.

src/renderer/envMapCreation.js
lyonsno: > `angleBetweenSphericals()` is still creating two JS objects every call. Did you measure the performance difference of caching those objects?
> 
> This is starting to look good 👍

Thanks :D
I could have sworn that I did, but apparently not as there is a significant benefit to avoiding the re-declaration!  in the softness=0.0 case it only saves like 8ms, but for softness=1.0 where we iterate over every pixel it goes from 1.4s to .75s!

I used a function expression to pull it off, but I'm very open to hearing other ideas. Weird scoping in javascript is not my specialty.

src/renderer/envMapCreation.js
lyonsno: > `angleBetweenSphericals()` is still creating two JS objects every call. Did you measure the performance difference of caching those objects?
> 
> This is starting to look good 👍

Thanks :D
I could have sworn that I did, but apparently not as there is a significant benefit to avoiding the re-declaration! in the softness=0.0 case it only saves like 8ms, but for softness=1.0 where we iterate over every pixel it goes from 1.4s to .75s!

I used a function expression to pull it off, but I'm very open to hearing other ideas. Weird scoping in javascript is not my specialty.

lyonsno:
> `angleBetweenSphericals()` is still creating two JS objects every call. Did you measure the performance difference of caching those objects?
> 
> This is starting to look good 👍

Thanks :D
I could have sworn that I did, but apparently not as there is a significant benefit to avoiding the re-declaration! in the softness=0.0 case it only saves like 8ms, but for softness=1.0 where we iterate over every pixel it goes from 1.4s to .75s!

I used a function expression to pull it off, but I'm very open to hearing other ideas. Weird scoping in javascript is not my specialty.

PR #19, opened 2019-09-27
Multiple render targets; Refactors

jaxry:
## Brief Description
This PR improves the RenderTarget class to allow for multiple render targets. This is useful for the denoising pipeline and hybrid rendering.

In addition, I've introduced refactors made in the [hybrid rendering branch](https://github.com/hoverinc/ray-tracing-renderer/tree/hybrid-rendering) that aren't actually related to hybrid rendering. This is so we can include these improvements in `master` without including the hybrid rendering stuff yet.

## Pull Request Guidelines


src/RayTracingRenderer.js
```diff
@@ -1,5 +1,5 @@
 import { loadExtensions } from './renderer/glUtil';
-import { makeRenderingPipeline } from './renderer/renderingPipeline';
+import { makeRenderingPipeline } from './renderer/RenderingPipeline';
```
jaxry: Every file which returns a `make[X]` factory function has been capitalized. This corresponds to the code style that classes begin with capital letters. Since the `make[X]` function isn't *really* a class but instead a factory function, we leave the `m` lower case. But the capital letter in the file name signifies that the file contains a factory function which is used in the same way as a class.

Thoughts?

src/renderer/Framebuffer.js
```diff
@@ -0,0 +1,105 @@
+import { makeTexture } from "./Texture";
```
jaxry: `renderTarget.js` has been renamed to `Framebuffer.js`. It now supports one or multiple render targets.

src/renderer/Framebuffer.js
```diff
@@ -0,0 +1,105 @@
...
+export function makeFramebuffer(params) {
+  const {
+    depth = false, // use depth buffer
+    gl,
+    linearFiltering = false, // linearly filter textures
+
+    // supply one but not both of these parameters
+    renderTarget, // single render target: { storage: 'byte' | 'float' }
+    renderTargets, // multiple render targets: RenderTargets object
+  } = params;
```
jaxry: Are these argument options confusing?

src/renderer/RenderTargets.js
```diff
@@ -0,0 +1,18 @@
+// targets is array of { name: string, storage: 'byte' | 'float'}
```
jaxry: Though unused for the time being, this factory function is intended to be instantiated by shader functions which output to multiple render targets.

src/renderer/TextureAllocator.js
```diff
@@ -19,8 +19,10 @@ export function makeTextureAllocator(gl) {
 
   function reserveSlot() {
     const unit = nextUnit++;
-    return (uniform, textureObj) => {
-      bindGl(uniform, textureObj, unit);
+    return {
+      bind(uniform, textureObj) {
+        bindGl(uniform, textureObj, unit);
+      }
```
jaxry: Returning an object with a method is less confusing to me than just returning a function.

src/renderer/glsl/chunks/bsdf.glsl
```diff
@@ -1,4 +1,4 @@
-export default function(params) {
+export default function(defines) {
```
jaxry: `defines` is more accurate than `params`

src/renderer/mergeMeshesToGeometry.js
```diff
@@ -10,7 +9,7 @@ export function mergeMeshesToGeometry(meshes) {
   const materialIndexMap = new Map();
 
   for (const mesh of meshes) {
-    const geometry = mesh.geometry.clone();
+    const geometry = cloneBufferGeometry(mesh.geometry, ['position', 'normal', 'uv']);
```
jaxry: `mesh.geometry.clone()` clones all attributes, even the ones we don't use. I've replaced this with our own method that only clones specific attributes. This improves performance by ~10 ms with standard meshes. Not much, but if the user provides an unusual mesh with lots of unused attributes, the original `clone()` will take a heavier performance hit.

src/renderer/mergeMeshesToGeometry.js
```diff
@@ -63,25 +63,65 @@ function mergeGeometry(geometryAndMaterialIndex, vertexCount, indexCount) {
...
+
+  const index = bufferGeometry.getIndex();
+  if (index) {
+    newGeometry.setIndex(index);
+  }
+
+  return newGeometry;
+}
+
+function addFlatGeometryIndices(geometry) {
```
jaxry: Move this function here and delete the standalone file. The function is only used once and this is the only place I can see it being used.

src/RayTracingRenderer.js
lyonsno: Works for me.

src/renderer/RenderingPipeline.js
```diff
@@ -34,8 +34,18 @@ export function makeRenderingPipeline({
 
   const useLinearFiltering = optionalExtensions.OES_texture_float_linear;
 
-  const hdrBuffer = makeRenderTargetFloat(gl); // full resolution buffer representing the rendered scene with HDR lighting
-  const hdrPreviewBuffer = makeRenderTargetFloat(gl, useLinearFiltering); // lower resolution buffer used for the first frame
+  // full resolution buffer representing the rendered scene with HDR lighting
+  const hdrBuffer = makeFramebuffer({
+    gl,
+    renderTarget: { storage: 'float' }
```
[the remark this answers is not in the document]
jaxry: This is really tricky. `makeRenderTargets` called with a single target doesn't make much sense. It's intended to store render targets with different names (hence the existence of the `glslOutput()` method). But passing it a single target without a name wouldn't fit with `RenderTarget`'s functionality.

What do you think about this idea:

* `makeFrameBuffer` takes a single `renderTargets` argument. It's either a single render target in `{ storage }` format, or multiple render targets as an instance of `RenderTargets`.

Maybe that's still clunky. Do you have other ideas?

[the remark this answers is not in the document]
lyonsno: That sounds like the best idea so far.  What about passing in an array of [ { name, storage }, { name, storage } ] or a { storage } object and then calling makeRenderTargets() on the array from inside makeFrameBuffer if there are multiple objects passed in?  That might make the single vs multiple arguments even more similar.  If you don't like that idea then I think what you've suggested here is good to merge.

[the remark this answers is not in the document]
jaxry: That's a clean way to do it, but I have to keep them separate for a reason (at least with this implementation). The shader itself needs to know the render target locations (as integer ids), and shaders are generally compiled before framebuffer initialization. So `RenderTargets` supplies locations to both the shader and the framebuffer. Creating the RenderTargets object in the framebuffer means we'd need to create the framebuffer before the shaders?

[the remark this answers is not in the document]
lyonsno: ah I see.  That doesn't sound like a worthwhile tradeoff.  Things could maybe be more elegant somehow but it's not confusing anymore as of now so let's go with it and we can always improve it later.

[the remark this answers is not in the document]
jaxry: I'll keep thinking about it, and the nicest solution will probably come up once we're using multiple buffers and noticing usage patterns.

PR #27, opened 2019-10-06
Tweak Directional Light

jaxry:
## Brief Description

* Add `softness` parameter to `SoftDirectionalLight`'s constructor. This makes softness similar to `intensity` and `color`, which makes sense to me since all three configure the appearance of the light, but only two were providential as constructor parameters before.

* Use exponent table for `rgbeToFloat`. This improves performance from 80 ms to 60 ms for that function (with a 4k envmap), and lets us use the original `Math.pow` function, which is less confusing to me than doing bit manipulation.

* Cleanup. Added whitespace for more clarity, and restructured code slightly.

## Pull Request Guidelines

### DIFF
--- a/src/renderer/envMapCreation.js
+++ b/src/renderer/envMapCreation.js
@@ -20,17 +20,21 @@ export function generateEnvMapFromSceneComponents(directionalLights, environment
 
 export function initializeEnvMap(environmentLights) {
   let envImage;
+
   // Initialize map from environment light if present
   if (environmentLights.length > 0) {
     // TODO: support multiple environment lights (what if they have different resolutions?)
     const environmentLight = environmentLights[0];
+
     envImage = {
       width: environmentLight.map.image.width,
       height: environmentLight.map.image.height,
       data: environmentLight.map.image.data,
     };
+
     envImage.data = rgbeToFloat(envImage.data, environmentLight.intensity);
-  } else { // initialize blank map
+  } else {
+    // initialize blank map
     envImage = generateBlankMap(DEFAULT_MAP_RESOLUTION.width, DEFAULT_MAP_RESOLUTION.height);
   }
 
@@ -40,7 +44,6 @@ export function initializeEnvMap(environmentLights) {
 export function generateBlankMap(width, height) {
   const texels = width * height;
   const floatBuffer = new Float32Array(texels * 3);
-  floatBuffer.fill(0.0);
 
   return {
     width: width,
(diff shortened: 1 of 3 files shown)


src/renderer/envMapCreation.js
```diff
@@ -63,58 +68,61 @@ function addLightAtCoordinates(light, image, originCoords) {
...
 
   // Iterates over each row from top to bottom
   for (let i = 0; i < xTexels; i++) {
+
+    let encounteredInThisRow = false;
+
     // Iterates over each texel in row
     for (let j = 0; j < yTexels; j++) {
-      bufferIndex = j * width + i;
+      const bufferIndex = j * width + i;
```
jaxry: Declared variables in the lowest scope they're used. Is there a reason to declare them higher? There's no difference in performance in JS.

jaxry:
Some of the style changes are opinionated. What do you think?

src/renderer/envMapCreation.js
lyonsno: I didn't actually measure performance difference there, so maybe my understanding is fuzzy.  I thought that since `let` and `const` were block-scope declarations they would end up declared millions of times if declared inside the loop?

src/renderer/rgbeToFloat.js
```diff
@@ -1,24 +1,26 @@
...
 // Optional multiplier argument for performance optimization
-export function rgbeToFloat(buffer, multiplier = 1) {
+export function rgbeToFloat(buffer, intensity = 1) {
   const texels = buffer.length / 4;
   const floatBuffer = new Float32Array(texels * 3);
 
+  const expTable = [];
+  for (let i = 0; i < 255; i++) {
+    expTable[i] = intensity * Math.pow(2, i - 128) / 255;
+  }
```
lyonsno: Nice 👍

src/renderer/rgbeToFloat.js
```diff
@@ -1,24 +1,26 @@
 // Convert image data from the RGBE format to a 32-bit floating point format
 // See https://www.cg.tuwien.ac.at/research/theses/matkovic/node84.html for a description of the RGBE format
 // Optional multiplier argument for performance optimization
-export function rgbeToFloat(buffer, multiplier = 1) {
+export function rgbeToFloat(buffer, intensity = 1) {
```
lyonsno: This felt a little gross to me when I wrote it, since it goes against the name of the function, but I don't have any better ideas.  I take it you're at least ok with it?

src/renderer/envMapCreation.js
jaxry: JS tends to optimize anything like that out of the equation. But to be sure I did benchmark before and after, and there was no measurable difference.

src/renderer/rgbeToFloat.js
jaxry: Yup. This seems like a nice way to do it.

Issue #20, opened 2019-09-30
ES6 module not working


ricardomatias:
Error: `export 'EnvironmentLight' (imported as 'THREE') was not found in 'three'`

It's not working with `npm install` using ES6 modules. As far as I see, your `main.js` is not being imported anywhere which doesn't expose the rest of the modules.

jaxry:
Thanks for the ticket. The npm package is definitely not configured correctly... It looks like the `main` field in the [package.json](https://github.com/hoverinc/ray-tracing-renderer/blob/master/package.json) isn't pointing to the right file. If you navigate to the dependency in your `node_modules` folder, and change that property to `src/main.js`, does the environment light import then?

If that takes care of it, would you like to submit a PR with the change? If it's too much trouble, I can take care of it all the same too 👍

jaxry:
I believe I've addressed the problem in the latest version. If it's still an issue, let me know and we can try figuring it out.

ricardomatias:
I'm getting the error `THREE is not defined` at `child instanceof THREE.EnvironmentLight` from `RayTracingShader.js`

I've actually spent quite a bit of time playing around with rollup's config and imports and I didn't manage to make it work. I suspect it has something to do with the way things are imported. There are examples of using the global/window `three` instance but there are also occasions where the import version is used.

Unrelated, the keyword `module` should also not be used (as in `RayTracingRenderer.js`) since it might create conflicts.

jaxry:
Thanks for looking into this. I'm not sure why you're getting that error from [RayTracingShader](https://github.com/hoverinc/ray-tracing-renderer/blob/master/src/renderer/RayTracingShader.js), since `three` is imported in that file. Rollup is configured so that three.js should be included in the HTML before the code that contains the ray tracing renderer. Are you doing that? If so, the issue is deeper.

>Unrelated, the keyword module should also not be used (as in RayTracingRenderer.js) since it might create conflicts.

That's a good point. Taking a peek at the [build file](https://github.com/hoverinc/ray-tracing-renderer/blob/master/build/RayTracingRenderer.js), and `module` is used by the build. But it's only used outside the ray tracing renderer's usage of the variable, and will be kept separate via scoping. Maybe it's a good idea to change it anyway? But I don't see an immediate problem.

elfrank:
Hi @ricardomatias , thanks for reporting this issue.

[...]

PR #29, opened 2019-10-09
Fix module importing

jaxry:
## Brief Description
Fixes the issue described in https://github.com/hoverinc/ray-tracing-renderer/issues/20. I've tested that the renderer now correctly loads via

* HTML-loaded Three.js and HTML-loaded Ray Tracing Renderer
* HTML-loaded Three.js and module-loaded Ray Tracing Renderer
* Module-loaded Three.js and module-loaded Ray Tracing Renderer

## Pull Request Guidelines

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -5,8 +5,8 @@ import { EnvironmentLight } from './EnvironmentLight';
 import { RayTracingMaterial } from './RayTracingMaterial';
 import { RayTracingRenderer } from './RayTracingRenderer';
 
-/* global THREE */
-if (THREE) {
+if (window.THREE) {
+  /* global THREE */
   THREE.LensCamera = LensCamera;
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -5,8 +5,8 @@ import { EnvironmentLight } from './EnvironmentLight';
 import { RayTracingMaterial } from './RayTracingMaterial';
 import { RayTracingRenderer } from './RayTracingRenderer';
 
-/* global THREE */
-if (THREE) {
+if (window.THREE) {
```
elfrank: @jaxry what's the difference (in practice)?

jaxry: If the `THREE` object doesn't exist, as is the case when you load three.js as a module, referencing `THREE` in the code (even in the `if`) will throw an exception. So we check it as a property on the window object instead.

PR #34, opened 2019-10-21
Use ES6 build for the main library entry point

bnjm:
## Brief Description
Firstly, thanks for the amazing project.

The current library entry point is `src/main.js`, meaning if you `npm install ray-tracing-renderer` then you'll import the untranspiled modules into your app. This almost works, but most bundlers won't recognize .vert or .frag files out of the box, in the case of wepback this results in a runtime error when trying to use the library.

Either it could be documented that glsl files need to be recognized by the consumers' bundler, or the entry could be pointed at the ES6 build as in this PR.

## Pull Request Guidelines

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "ray-tracing-renderer",
   "version": "0.2.0",
   "description": "A [Three.js](https://github.com/mrdoob/three.js/) renderer which utilizes path tracing to render a scene with true photorealism. The renderer supports global illumination, reflections, soft shadows, and realistic environment lighting.",
-  "main": "src/main.js",
+  "main": "build/RayTracingRenderer.js",
   "scripts": {
     "build": "./node_modules/.bin/rollup -c",
     "dev": "./node_modules/.bin/rollup -cw --environment DEV & ./node_modules/.bin/http-server",


jaxry:
Is it typical for a package to expose the UMD module instead of the original source files? If it is then we should do that regardless. 

But if it's up for debate, then should we even be importing `vert` and `frag` files? If you take a look at them, you'll see they're actually minimal JS functions that return a template string. So strictly speaking those are actually `js` files. I renamed them to so that code editors apply syntax highlighting to the GLSL bit. Do you think it makes more sense to save them as `js` files and give up the automatic syntax highlighting? And if that's the case, would it make sense for the package to use the UMD module or `main.js`?

bnjm:
> Does it make more sense to expose the UMD module instead of the original source files in general? If so, then we should do this regardless.
> 

That seems to be what's commonly done when a build step is needed.

> But if it's up for debate, then should we even be importing `vert` and `frag` files? If you take a look at them, you'll see they're actually minimal JS functions that return a template string. So strictly speaking those are actually `js` files. I renamed them to so that code editors apply syntax highlighting to the GLSL bit. Do you think it makes more sense to save them as `js` files and give up the automatic syntax highlighting? And if that's the case, would it make sense for the package to use the UMD module or `main.js`?

If you don't want to lose the syntax highlighting what probably makes sense is to create an additional ESM build, and add that as the `module` entry in `package.json`. Tools like webpack and rollup will use the `module` entry point, and it could potentially lead to a smaller bundle with tree-shaking support. I'm happy to try that out and add it to this PR or another. 

If you don't mind losing the `vert` and `frag` files then you could just point the `module` entry directly at `src/main.js`.

elfrank:
@bnjm thanks! It all sounds reasonable and additional PRs are welcomed.

It also looks like the Github action is not working with forks and that's on me to investigate.

Merging for now and we'll figure out the rest in separate PRs.

@jaxry feel free to give your two cents on how to move forward with `glsl` code in the repo.

jaxry:
At this point there's nothing optional to tree shake, so importing the bundle should be good for now. 👍

PR #36, opened 2019-10-26
chore(sample-scene): include sample scene on the repo

elfrank:
## Brief Description
This PR includes the sample scene as part of the repo.

* Sample scene that shows material support on the renderer is now part of the repo located under the scenes/ directory
* Modified `.gitignore` file such that any other folder inside of `scenes/` would be ignored by git.
* includes `.npmignore` for the first time. When a developer includes the renderer as an npm package, the dependency won't include the `scenes` folder.

Screen Shot 2019-10-26 at 2 08 27 PM

## Pull Request Guidelines


jaxry:
I pushed the version of `main.js` that I've been using. The scene is rotated by 180 degrees to be compatible with the envmap rotation PR we made a few weeks ago.

.gitignore
```diff
@@ -1,10 +1,18 @@
...
 
 # demo/test scenes
 demos/
 test-scene/
+
+# ignore everything on the "scenes/" directory
+scenes/*
+
+# except for "scenes/materials"
+!scenes/materials
```
jaxry: For the sake of simplicity, what about omitting all these `demos/`, `test-scene/`, `scene/` etc. ignores? If there are directories we're using that we'd rather not commit, I find it's easy to just leave the directories untracked. 

`.gitignore` is handy for automatically generated build files that are always generated, but ignoring all these directories sounds like a rabbit hole we could avoid.

elfrank:
> I pushed the version of `main.js` that I've been using. The scene is rotated by 180 degrees to be compatible with the envmap rotation PR we made a few weeks ago.

To what PR are you referring to @jaxry ? This one: https://github.com/hoverinc/ray-tracing-renderer/pull/3 ?
As far as I remember, we didn't merge changes related to rotating environment maps. Perhaps it was something else and it's skipping my mind?

jaxry:
This one https://github.com/hoverinc/ray-tracing-renderer/pull/4

.gitignore
jaxry: Git leaves any new files untracked and they won't be committed unless the user explicitly adds them via `git add`. I think simply leaving extraneous scenes untracked is a perfectly good alternative.

`.gitignore` is useful if there are certain files you know you want to exclude that are present on everybody's clone. i.e. `node_modules`, but for personal files like the ones we're dealing with, leaving them untracked and not adding them with `git add` seems sufficient to me.

elfrank: done, could you take a look again @jaxry ?

elfrank:
> Looks good, although I should point out one thing. This example has `renderer.maxHardwareUsage` set to `true` which makes it unsuitable for a general purpose demo since it will lock the computer at full resolution. The scene is intended to test the accuracy and performance of the renderer. With that in mind, would calling the directory `scene/renderer-test` (or something along those lines) be more precise? I don't think `materials` is a problem; I just don't want anyone getting the wrong idea.

sounds reasonable, I just updated the example.

PR #39, opened 2019-10-30
Temporal Filtering (reverse reprojection)

jaxry:
## Brief Description
Progress:
- [x] Add shader pass which reprojects previous camera angles
- [x] Reject history if surface normals don't match
- [x] Smoothly blend between reprojected "imperfect" render, and a fresh render without artifacts
- [x] Reduce flickering from overly sensitive history rejection
- [x] Improve history rejection by considering position history
- [x] Add support for anti-aliasing
- [x] Incorporate existing pipeline (using low res preview image)
- [x] Improve performance by only reprojecting once when camera moves
- [ ] Test performance on a variety of devices
- [ ] ~~Add support for Lens Camera~~ To implement in future PR

TODO

## Pull Request Guidelines


src/RayTracingRenderer.js
```diff
@@ -4,6 +4,7 @@ import * as THREE from 'three';
 
 const glRequiredExtensions = [
   'EXT_color_buffer_float', // enables rendering to float buffers
+  'EXT_float_blend',
```
jaxry: This should have always been here. It was only a recent version of Firefox that started to warn this extension was necessary, and that it was turning it on implicitly.

elfrank: @jaxry interestingly, it doesn't show up in webglstats: https://webglstats.com/
any thoughts on that?

carlos-lopez-garces:
I've been studying the PR for the past 2 evenings. I think you can expect to see questions and comments from me in the next few days.

carlos-lopez-garces:
I ran the renderer on the test scene (spheres of different materials) on 2 devices (shared the videos over Slack):

```
CPU: 2.6 GHz Intel Core i7. RAM 16 GB.
Graphics: Radeon Pro Vega 16 4 GB Intel UHD Graphics 630 1536 MB.
OS: macOS Mojave.

CPU: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz, 2601 Mhz, 6 Core(s), 12 Logical Processor(s). Graphics: NVIDIA GeForce RTX 2060. 
OS: Windows 10 Home.
```

The render on macOS+Radeon looked virtually the same as the video shared by @jaxry; the same characteristics: the smearing of the floor disappears fast when rotating (< 1s); only translucent materials display rather big fireflies at the beginning, but they reduce in size quickly; the smearing looks smooth.

The render on Windows+GeForce also looked smooth. Perhaps the only thing I noticed was that 1) the initial load time was higher, and 2) the smearing of the floor remains longer when rotating.

src/renderer/RenderingPipeline.js
```diff
@@ -32,25 +45,42 @@ export function makeRenderingPipeline({
...
+    renderTarget: rayTracingRenderTargets
+  });
+
+  let reprojectPreviewBuffer = makeFramebuffer({
+    gl,
+    renderTarget: rayTracingRenderTargets,
+    linearFiltering: true
+  });
+
+  let lastUsedBuffer = reprojectPreviewBuffer;
```
carlos-lopez-garces: 2 distinct sets of buffers are used: Those that are fed to `reproject.frag` as `hdrBuffer`s (i.e. `hdrPreviewBuffer` and `hdrBuffer` here) and those that are fed to it as `historyBuffer`s (i.e. `reprojectPreviewBuffer` and `reprojectBuffer` here).

The name `lastUsedBuffer` then become ambiguous, as one may think that it points to a buffer of either set. But in reality, `lastUsedBuffer` only ever points at `reprojectPreviewBuffer` or `reprojectBuffer`. 

A better name for it would then be `lastUsedReprojectBuffer`.

src/renderer/RenderingPipeline.js
```diff
@@ -153,77 +181,127 @@ export function makeRenderingPipeline({
...
+    }
+
+    if (!areCamerasEqual(camera, lastCamera)) {
+      initFirstSample();
       setPreviewBufferDimensions();
-      renderPreview();
+
+      rayTracingShader.setCamera(camera);
+      updateSeed(hdrPreviewBuffer.width, hdrPreviewBuffer.height);
+      newSampleToBuffer(hdrPreviewBuffer);
```
carlos-lopez-garces: The relationship between `historyBuffer`, `reprojectPreviewBuffer`, `reprojectBuffer` and `lastUsedBuffer`, the roles they play, and how essential the 4 are to the solution is clear to me, I believe.

What's not as clear is the need for `hdrPreviewBuffer`. It seems to me that we can do with `hdrBuffer` only; I even replaced the occurrences of `hdrPreviewBuffer` with `hdrBuffer`, rebuilt, and ran the test scene, and things continued to look the same.

My understanding is this:

- `hdrPreviewBuffer` stores the results of the 1st `rayTrace draw call / shader run` that takes place after a camera move, and serves as the `hdrBuffer` of the 1st `reprojection`.

- `hdrBuffer` gets cleared upon every camera move, and then stores the results of the subsequent `rayTrace draw calls / shader runs` of all of the tiles, to finally serve as `hdrBuffer` of the `reprojection` when the last tile is reached.

- Either `reprojectPreviewBuffer` or `reprojectBuffer ` are the buffers that get presented to the screen most of the time. The only exception is when `isLastTile && blendAmount <= 0.0`, in which case it is `hdrBuffer`.

Since `hdrBuffer` is practically an internal buffer that doesn't get presented to the screen (exception noted above),  cannot we also use `hdrBuffer` when the camera just moved, to store the results of the 1st `rayTrace` call? (Had `hdrBuffer` been presented to the screen, the concurrent write made by the 1st post-move `rayTrace` call would have messed up the screen, but it wasn't.)

I guess the answer is that a post-camera-move draw call that follows a `isLastTile && blendAmount <= 0.0` draw call, and that writes to `hdrBuffer` (instead of `hdrPreviewBuffer`, as it does now)  would mess up the screen because `hdrBuffer` was presented. (?)

jaxry: It's definitely possible to reduce the number of buffers, and that's an incredible important thing to do for memory optimization. I'm planning on tackling this as my next PR. 

> It seems to me that we can do with hdrBuffer only; I even replaced the occurrences of hdrPreviewBuffer with hdrBuffer, rebuilt, and ran the test scene, and things continued to look the same.

I'm not able to reproduce this Did you get this working with `maxHardwareUsage` set to false? When I try this, I get the following problem: `hdrPreviewBuffer` and `hdrBuffer` are different resolutions, and if we replace the preview buffer with the full screen buffer, we would be erroneously rendering at full res when we want to be rendering at a reduced res. The way around this would be to constrain the `glViewport` to a fraction of the full resolution buffer. But if we do this, we need to also scale the texture lookup coordinates the same amount in any shader reading a buffer rendered with a smaller-than-full viewport.

That's a mouthful but that's what I'm planning on doing with my next PR. Could you go into more detail if you managed to get this working already?

src/renderer/RenderingPipeline.js
jaxry: Aside from the two buffers you mentioned, lastUsedBuffer will also sometimes be `hdrBuffer`. Does `lastToneMappedBuffer` sound like a better name? The idea is that it doesn't matter what those buffers are, just that it was the last buffer rendered to the screen. That's the buffer we want to use to reproject to the new camera angle, since it's whatever buffer contains the "final" state of the render.

src/renderer/RenderingPipeline.js
```diff
@@ -153,77 +181,127 @@ export function makeRenderingPipeline({
...
-        updateSeed();
+        updateSeed(hdrBuffer.width, hdrBuffer.height);
       }
 
-      renderTile(x, y, tileWidth, tileHeight);
+      renderTile(hdrBuffer, x, y, tileWidth, tileHeight);
 
       if (isLastTile) {
-        hdrBufferToScreen();
+        let blendAmount = clamp(1.0 - sampleCount / maxReprojectedSamples, 0, 1);
```
carlos-lopez-garces: I see that `blendAmount` gets clamped between 0 and 1 here, then squared, and then clamped between 0 and 1 again inside [`reprojectShader.setBlendAmount`](https://github.com/hoverinc/ray-tracing-renderer/blob/reprojection/src/renderer/ReprojectShader.js#L40):

```js
function setBlendAmount(x) {
  gl.useProgram(program);
  gl.uniform1f(uniforms.blendAmount, clamp(x, 0, 1));
}
```

The square of the [0-1]-clamped number will still lie between 0 and 1, so the second clamping might be unnecessary.

Looking at the other 2 `setBlendAmount` calls:

- The one in [`drawFull`](https://github.com/hoverinc/ray-tracing-renderer/blob/reprojection/src/renderer/RenderingPipeline.js#L271-L273) is the same case.

- The other one made by [`drawTile`](https://github.com/hoverinc/ray-tracing-renderer/blob/reprojection/src/renderer/RenderingPipeline.js#L198) when the camera changes doesn't pass the input clamped, but its value (the non-configurable constant [`reprojectDecay`](https://github.com/hoverinc/ray-tracing-renderer/blob/reprojection/src/renderer/RenderingPipeline.js#L26)) already lies between 0 and 1.

If we wanted to avoid the double clamping, I'm not so sure which `clamp` call we should remove:

- On the one hand, blend factors are always values between 0 and 1, so it makes sense for `setBlendAmount` to enforce that by clamping the received input.

- On the other hand, the [`if (blendAmount > 0.0)`](https://github.com/hoverinc/ray-tracing-renderer/blob/reprojection/src/renderer/RenderingPipeline.js#L228) conditional below might be satisfied spuriously  if the unclamped `1.0 - sampleCount / maxReprojectedSamples` expression is negative, because the squaring will turn it positive and > 0.0.

src/renderer/glsl/rayTrace.frag
```diff
@@ -172,13 +172,15 @@ vec4 integrator(inout Ray ray) {
...
+  bounce(path, 1, si);
+
   // Manually unroll for loop.
   // Some hardware fails to interate over a GLSL loop, so we provide this workaround
-
-  ${unrollLoop('i', 1, defines.BOUNCES + 1, 1, `
+  SurfaceInteraction indirectSi;
+  ${unrollLoop('i', 2, defines.BOUNCES + 1, 1, `
   // equivelant to
   // for (int i = 1; i < defines.bounces + 1, i += 1)
```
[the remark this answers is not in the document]
jaxry: Good catch!

jaxry:
I made tweaks to the code according to your comments.

src/renderer/RenderingPipeline.js
carlos-lopez-garces: Let me check again and I'll get back to you.

carlos-lopez-garces: I see what you are saying, to some extent. Upon looking closer at the rendered images, I see that they become very blurry and low-res with the change I made (i.e. naively referring to `hdrBuffer` whenever `hdrPreviewBuffer` was used).

Change vs no change:

image

image

Please disregard.

PR #41, opened 2019-11-08
Hybrid rendering

lyonsno:
Just what I've got so far.  Next gonna look at anti-aliasing the g-buffer

## Pull Request Guidelines


src/renderer/glsl/multiply.frag
```diff
@@ -0,0 +1,58 @@
...
+  }
+  light = pow(light, vec3(1.0 / 2.2));
+  // vec3 light = texB.rgb;
+  // light = pow(light, vec3(1.0 / 2.2)); // gamma correction
+
+  fragColor = vec4(light, 1.0);
+}
+
+`;
+}
```
jaxry: Are we able to just do this in toneMap.frag?

src/renderer/glsl/gBuffer.frag
```diff
@@ -0,0 +1,64 @@
...
+} gBufferMaterials;
+
+#ifdef NUM_DIFFUSE_MAPS
+  uniform mediump sampler2DArray diffuseMap;
+#endif
+
+#ifdef NUM_NORMAL_MAPS
+  uniform mediump sampler2DArray normalMap;
+#endif
+
```
jaxry: Once we know this is how we want to do things, this is one of many bits of copy-pasted code we should extract into its own chunk/function. 👍

src/renderer/GBufferShader.js
```diff
@@ -0,0 +1,169 @@
...
+    NUM_MATERIALS: materials.length,
+    NUM_DIFFUSE_MAPS: maps.map.textures.length,
+    NUM_NORMAL_MAPS: maps.normalMap.textures.length,
+  }));
+  const program = createProgram(gl, vertShader, fragShader);
+
+  gl.useProgram(program);
+
+  const attributes = getAttributes(gl, program);
+  const uniforms = getUniforms(gl, program);
```
jaxry: This pattern of compiling shaders and getting the attributes and uniforms is shared in every shader file. We should find a way to abstract and solidify this design.

jaxry:
Starting to look good! G-buffers are much, much nicer to deal with (and it conserves memory & texture units) if each buffer is part of the same Array Texture. That's how we're currently doing it in the temporal filtering branch. When you have some time, could you see how I do it there and let me know what you think?

src/renderer/GBufferShader.js
lyonsno: oh yeah of course.  I'm just rolling in the mud here for now.

src/renderer/glsl/rayTrace.frag
```diff
@@ -113,40 +114,73 @@ ${sampleShadowCatcher(defines)}
...
       if (si.materialType == SHADOW_CATCHER) {
-        path.li += sampleShadowCatcher(si, i, path.ray, path.beta, path.alpha, path.li, path.abort);
+        vec3 newSample = sampleShadowCatcher(si, i, path.ray, path.beta, path.alpha, path.li, path.abort);
+        if (i <= 1) {
+          newSample /= max(si.color, vec3(0.001));
+          path.primaryAlpha = path.alpha;
+          path.primaryLi = newSample;
+        } else {
+          path.secondaryLi += newSample;
+          path.secondaryAlpha = path.alpha;
```
jaxry: Yep, this looks right to me! When you get to refactoring this code to be cleaner, I think we can do a few things:
* The BRDF function could omit the color altogether, so you can multiply it in for secondary bounces, instead of dividing it out here. That seems simpler to understand with a possible performance boost?

* These `if` checks could be moved outside to a single `if` check in `integrator()`

* The blending and tonemapping step could all be done in a single, lean shader call. In that shader we can optimize what you already have a bit to omit branching (using `mix`es and `step`s instead of `if`s. A shader with zero branching would likely give us even more performance)

But yeah, just the ideas I had looking at this WIP. What else are you thinking about at this point?

lyonsno: > Yep, this looks right to me! When you get to refactoring this code to be cleaner, I think we can do a few things:
> 
> * The BRDF function could omit the color altogether, so you can multiply it in for secondary bounces, instead of dividing it out here. That seems simpler to understand with a possible performance boost?

What do you think about eventually keeping hybrid as a flag vs adopting it as our only first bounce technique?  Either way I'm probably with you here.

> * These `if` checks could be moved outside to a single `if` check in `integrator()`

yup :)

> * The blending and tonemapping step could all be done in a single, lean shader call. In that shader we can optimize what you already have a bit to omit branching (using `mix`es and `step`s instead of `if`s. A shader with zero branching would likely give us even more performance)

I agree with this too, I guess I'm a little afraid of scope-creep vs our current naming conventions.  At the moment though I agree that it doesn't make a ton of logical sense to keep these seperate.  Only if we wanted intermediary processing steps between combining and re-modulating lighting and tone mapping.

> 
> But yeah, these are just the ideas I had looking at this WIP. What else are you thinking about at this point?

I'm pretty happy with things overall.  I have a slack q for you re: rendertargets in our slack channel.  Currently there are some discrepancies between the hybrid branch and the example scene, but I think those may all be resolved once the aliasing issue is.  There's also an issue where shading seems to show uv seams for a moment or two until it begins to converge, possibly only for secondary bounces.  Ultimately convergence doesn't seem to be impacted though, so I don't think it's introducing bias.

jaxry:
Closing this PR so we can focus on the new version

PR #42, opened 2019-11-25
chore(sample-models): create remote example

elfrank:
## Brief Description
This change updates `sample-models/index.html` to fetch resources from remote CDNs instead of using local file paths. This way, we can take advantage of using the example directly from the following URL: https://hoverinc.github.io/ray-tracing-renderer/scenes/sample-models/

There is an HTML clone, named `dev.html` for local development. Ideally, we would have a script that injects the paths accordingly. We can tackle that change in the future.

## Pull Request Guidelines

    - [x] I have updated the documentation accordingly.

### DIFF
--- a/scenes/sample-models/dev.html
+++ b/scenes/sample-models/dev.html
@@ -27,6 +27,6 @@
 </head>
 
 <body>
-  <script src="main.js"></script>
+  <script src="/scenes/sample-models/main.js"></script>
 </body>
 </html>
--- a/scenes/sample-models/dev.html
+++ b/scenes/sample-models/dev.html
@@ -1,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Path Tracing - Sample Models</title>
+
+  <script src="/node_modules/three/build/three.js"></script>
+  <script src="/node_modules/three/examples/js/loaders/GLTFLoader.js"></script>
+  <script src="/node_modules/three/examples/js/controls/OrbitControls.js"></script>
+  <script src="/node_modules/three/examples/js/loaders/RGBELoader.js"></script>
+  <script src="/node_modules/stats.js/build/stats.min.js"></script>
+  <script src="/node_modules/dat.gui/build/dat.gui.min.js"></script>
+  <script src="/build/RayTracingRenderer.js"></script>
+
+  <style>
+    html, body {
+      margin: 0;
+      padding: 0;
+      width: 100%;
+      height: 100%;
+    }
+
+    canvas {
+      display: block;
+    }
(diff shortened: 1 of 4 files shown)


scenes/sample-models/index.html
```diff
@@ -3,13 +3,15 @@
...
-  <script src="/node_modules/dat.gui/build/dat.gui.min.js"></script>
-  <script src="/build/RayTracingRenderer.js"></script>
+
+  <script src="https://cdn.jsdelivr.net/npm/three@0.108.0/build/three.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/three@0.108.0/examples/js/loaders/GLTFLoader.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/three@0.108.0/examples/js/controls/OrbitControls.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/three@0.108.0/examples/js/loaders/RGBELoader.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/three@0.108.0/examples/js/libs/stats.min.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/three@0.108.0/examples/js/libs/dat.gui.min.js"></script>
+  <script src="https://hoverinc.github.io/ray-tracing-renderer/build/RayTracingRenderer.js"></script>
```
jaxry: I think we could use a relative URL here as well, so that the repo isn't dependent on the name of the Github user. If we go down that path, the url `src="/ray-tracing-renderer/build/RayTracingRenderer.js"` should work. Would that make it easier to fork the repo?

Either way, this looks good!

scenes/sample-models/index.html
```diff
@@ -25,6 +27,6 @@
 </head>
 
 <body>
-  <script src="main.js"></script>
+  <script src="https://hoverinc.github.io/ray-tracing-renderer/scenes/sample-models/main.js"></script>
```
jaxry: Same change to possibly make here.

scenes/sample-models/index.html
elfrank: good idea

scenes/sample-models/index.html
elfrank: 👍

scenes/sample-models/dev.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Path Tracing - Sample Models</title>
+
+  <script src="/node_modules/three/build/three.js"></script>
```
elfrank: @jaxry now the path to three.js is the only difference between `dev.html` and `index.html`

PR #48, opened 2019-12-05
Scenes with a background property will by default display that as the…

lyonsno:
## Brief Description
de-couple scene lighting map from scene background

Scenes with a background property will by default display that as the background instead of any environment lights that might be present.
If no background property is specified scene will fall back to using an environment light for the background

An environment light with no map will now respect it's `color` property

## Pull Request Guidelines

### DIFF
--- a/src/renderer/envMapCreation.js
+++ b/src/renderer/envMapCreation.js
@@ -16,7 +16,7 @@ export function generateBackgroundMapFromSceneBackground(background) {
   let backgroundImage;
 
   if (background.isColor) {
-    backgroundImage = generateSolidMap(DEFAULT_MAP_RESOLUTION.width, DEFAULT_MAP_RESOLUTION.height, background);
+    backgroundImage = generateSolidMap(1, 1, background);
   } else if (background.encoding === THREE.RGBEEncoding) {
       backgroundImage = {
         width: background.image.width,
@@ -30,7 +30,7 @@ export function generateBackgroundMapFromSceneBackground(background) {
 
 export function generateEnvMapFromSceneComponents(directionalLights, ambientLights, environmentLights) {
   let envImage = initializeEnvMap(environmentLights);
-  ambientLights.forEach( light => { addAmbientLightToEnvMap(light, envImage) });
+  ambientLights.forEach( light => { addAmbientLightToEnvMap(light, envImage); });
   directionalLights.forEach( light => { envImage.data = addDirectionalLightToEnvMap(light, envImage); });
 
   return envImage;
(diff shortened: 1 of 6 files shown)


lyonsno:
> I believe the third variable can just be a `float` instead of a `vec3`. Does that work? And also, it might be worth measuring performance using a ternary operator, because that would be more straightforward if it performs the same.

Good call, float works fine.  The granularity of our performance testing leaves a little to be desired, but it actually seems like performance is slightly better with a ternary? Does that make any sense to you?

src/renderer/envMapCreation.js
```diff
@@ -25,33 +25,72 @@ export function initializeEnvMap(environmentLights) {
...
+    if (environmentLight.map && (environmentLight.map.encoding === THREE.RGBEEncoding || environmentLight.map.encoding === THREE.LinearEncoding)) {
+      envImage = {
+        width: environmentLight.map.image.width,
+        height: environmentLight.map.image.height,
+        data: environmentLight.map.image.data,
+      };
+      envImage.data = rgbeToFloat(envImage.data, environmentLight.intensity);
+    }
+    // if there is no map fall back to color property
+    else if (environmentLight.color && environmentLight.color.isColor) {
```
jaxry: Should this be 
```javascript
else if (environmentLight.isColor) `
```

src/renderer/envMapCreation.js
```diff
@@ -25,33 +25,72 @@ export function initializeEnvMap(environmentLights) {
...
+      envImage = generateMap(DEFAULT_MAP_RESOLUTION.width, DEFAULT_MAP_RESOLUTION.height, environmentLight.color, environmentLight.intensity)
```
jaxry: Does a `color` have an intensity?

jaxry: Since we're working with a solid color, can we just initialize a 1x1 texture with that color?

jaxry: Oh I see. I was confused! This PR also adds a `color` property to the EnvironmentLight, so we can illuminate the scene with a constant color. Instead of doing it this way, what about maintaining parity with Three.js and illuminating scenes with constant colors via `Three.AmbientLight`?

lyonsno: Thats actually how I was planning to write it initially!  My concern was that the code changes involved there were a bit more involved and that we might be re-writing some of this soon anyway.  I'm not totally apposed to going in and moving this functionality to an ambient light though, what do you think?

lyonsno: yup, good catch.

PR #50, opened 2019-12-06
Improve performance scaling

jaxry:
## Brief Description
This PR improves the algorithm that determines each draw call's tile size. This ensures the renderer remains responsive while also rendering as quickly as possible.

## Pull Request Guidelines

### DIFF
--- a/src/RayTracingRenderer.js
+++ b/src/RayTracingRenderer.js
@@ -27,7 +27,6 @@ export function RayTracingRenderer(params = {}) {
 
   let pipeline = null;
   const size = new THREE.Vector2();
-  let renderTime = 22;
   let pixelRatio = 1;
 
   const module = {
@@ -62,7 +61,6 @@ export function RayTracingRenderer(params = {}) {
       }
     };
 
-    module.setRenderTime(renderTime);
     module.setSize(size.width, size.height);
     module.needsUpdate = false;
   }
@@ -105,17 +103,6 @@ export function RayTracingRenderer(params = {}) {
   };
 
   module.getPixelRatio = () => pixelRatio;
-
-  module.setRenderTime = (time) => {
-    renderTime = time;
-    if (pipeline) {
-      pipeline.setRenderTime(time);
-    }
-  };
-
-  module.getRenderTime = () => {
-    return renderTime;
-  };
 
   module.getTotalSamplesRendered = () => {
     if (pipeline) {
--- a/src/renderer/RenderingPipeline.js
+++ b/src/renderer/RenderingPipeline.js
@@ -80,7 +80,7 @@ export function makeRenderingPipeline({
   }
 
   function setPreviewBufferDimensions() {
-    const desiredTimeForPreview = 8; // 60 fps
+    const desiredTimeForPreview = 8;
     const numPixelsForPreview = desiredTimeForPreview / tileRender.getTimePerPixel();
 
     const aspectRatio = hdrBuffer.width / hdrBuffer.height;
--- a/src/renderer/RenderingPipeline.js
+++ b/src/renderer/RenderingPipeline.js
@@ -80,15 +80,15 @@ export function makeRenderingPipeline({
   }
 
   function setPreviewBufferDimensions() {
-    const aspectRatio = hdrBuffer.width / hdrBuffer.height;
     const desiredTimeForPreview = 8; // 60 fps
     const numPixelsForPreview = desiredTimeForPreview / tileRender.getTimePerPixel();
+
+    const aspectRatio = hdrBuffer.width / hdrBuffer.height;
     const previewWidth = Math.round(clamp(Math.sqrt(numPixelsForPreview * aspectRatio), 1, hdrBuffer.width));
     const previewHeight = Math.round(clamp(previewWidth / aspectRatio, 1, hdrBuffer.height));
 
     const diff = Math.abs(previewWidth - hdrPreviewBuffer.width) / previewWidth;
-    if (diff > 0.1) {
-      console.log('NEW WIDTH', hdrPreviewBuffer.width, previewWidth);
+    if (diff > 0.05) { // don't bother resizing if the buffer size is only slightly different
       hdrPreviewBuffer.setSize(previewWidth, previewHeight);
     }
   }
--- a/src/renderer/RenderingPipeline.js
+++ b/src/renderer/RenderingPipeline.js
@@ -81,11 +81,14 @@ export function makeRenderingPipeline({
 
   function setPreviewBufferDimensions() {
     const aspectRatio = hdrBuffer.width / hdrBuffer.height;
-    const desiredTimeForPreview = 16; // 60 fps
+    const desiredTimeForPreview = 8; // 60 fps
     const numPixelsForPreview = desiredTimeForPreview / tileRender.getTimePerPixel();
-    const previewWidth = clamp(Math.sqrt(numPixelsForPreview * aspectRatio), 1, hdrBuffer.width);
-    const previewHeight = clamp(previewWidth / aspectRatio, 1, hdrBuffer.height);
-    if (previewWidth !== hdrPreviewBuffer.width) {
+    const previewWidth = Math.round(clamp(Math.sqrt(numPixelsForPreview * aspectRatio), 1, hdrBuffer.width));
+    const previewHeight = Math.round(clamp(previewWidth / aspectRatio, 1, hdrBuffer.height));
+
+    const diff = Math.abs(previewWidth - hdrPreviewBuffer.width) / previewWidth;
+    if (diff > 0.1) {
+      console.log('NEW WIDTH', hdrPreviewBuffer.width, previewWidth);
       hdrPreviewBuffer.setSize(previewWidth, previewHeight);
     }
   }
@@ -221,7 +224,6 @@ export function makeRenderingPipeline({
     drawOffscreenTile,
     drawFull,
     restartTimer: tileRender.restartTimer,
-    setRenderTime: tileRender.setRenderTime,
     setSize,
     hdrBufferToScreen,
     getTotalSamplesRendered() {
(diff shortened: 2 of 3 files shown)


src/RayTracingRenderer.js
```diff
@@ -106,17 +104,6 @@ export function RayTracingRenderer(params = {}) {
 
   module.getPixelRatio = () => pixelRatio;
 
-  module.setRenderTime = (time) => {
```
jaxry: I think we can simply remove this method from the API. It seems there's an ideal render time (~50 fps) which balances responsiveness and render speed that's the same between all the devices I tested on.

src/renderer/RenderingPipeline.js
```diff
@@ -80,12 +80,15 @@ export function makeRenderingPipeline({
...
-    const previewWidth = clamp(Math.sqrt(numPixelsForPreview * aspectRatio), 1, hdrBuffer.width);
-    const previewHeight = clamp(previewWidth / aspectRatio, 1, hdrBuffer.height);
-    if (previewWidth !== hdrPreviewBuffer.width) {
+
+    const aspectRatio = hdrBuffer.width / hdrBuffer.height;
+    const previewWidth = Math.round(clamp(Math.sqrt(numPixelsForPreview * aspectRatio), 1, hdrBuffer.width));
+    const previewHeight = Math.round(clamp(previewWidth / aspectRatio, 1, hdrBuffer.height));
+
+    const diff = Math.abs(previewWidth - hdrPreviewBuffer.width) / previewWidth;
+    if (diff > 0.05) { // don't bother resizing if the buffer size is only slightly different
```
jaxry: Resizing the buffer takes time and we can avoid it if the change if buffer size is small. This approach is a definite improvement, but I'm not happy with it altogether. 

In a separate PR, I will overhaul the buffer system so that the preview buffer actually uses a small viewport of the main buffer, instead of constantly resizing an independent buffer.

src/renderer/TileRender.js
```diff
@@ -81,14 +67,14 @@ export function makeTileRender(gl) {
...
       const timePerTile = timeElapsed / numTiles;
-      const error = desiredTimePerTile - timePerTile;
 
-      // higher number means framerate converges to targetRenderTime faster
-      // if set too high, the framerate fluctuates rapidly with small variations in frame-by-frame performance
-      const convergenceStrength = 1000;
+      const expAvg = 0.5;
+
+      const newPixelsPerTile = pixelsPerTile * desiredTimePerTile / timePerTile;
+      pixelsPerTile = expAvg * pixelsPerTile + (1 - expAvg) * newPixelsPerTile;
```
jaxry: Calculating the ideal pixels per tile (to match `desiredTimePerTile`) in a different way. This method is way faster in both converging to the ideal time, and also remaining stable once it reaches that time.

src/renderer/TileRender.js
```diff
@@ -120,32 +106,24 @@ export function makeTileRender(gl) {
...
-  } else if (maxRenderbufferSize === 16384 && maxViewportDims[0] >= 32768) {
-    return 100000;
-  } else if (maxRenderbufferSize >= 32768) {
     return 200000;
-  } else {
-    return 50000;
+  } else if (maxRenderbufferSize === 16384) {
+    return 400000;
+  } else if (maxRenderbufferSize >= 32768) {
+    return 600000;
```
jaxry: Significantly increase the initial estimates for all devices. The previous numbers were too low in practice and took a while to converge to the actual numbers for each device.

In the case that these numbers are too high, the new converging method quickly brings them to the right range (only 1 or 2 draw calls needed)

jaxry:
This PR brings a significant improvement for my laptop and desktop. We should test these changes on a few other devices too.

src/RayTracingRenderer.js
lyonsno: @jaxry which devices have you tested on?

jaxry: My Macbook with an integrated Intel GPU (mid end), and my desktop with the GTX 2080 (very high end) on both Linux and Windows (different drivers, different results). Both devices had the performance issues we've noted, and this PR fixes it for both. I think the most useful device to test at this point would be a low-end one.

lyonsno:
How low end?  Would a 13 inch mbp do the trick?

jaxry:
I think so. We should pay attention to whether the preview window lags when moving the camera, and whether the renderer remains responsive when the camera stays still.

src/renderer/RenderingPipeline.js
```diff
@@ -80,12 +80,15 @@ export function makeRenderingPipeline({
   }
 
   function setPreviewBufferDimensions() {
-    const aspectRatio = hdrBuffer.width / hdrBuffer.height;
-    const desiredTimeForPreview = 16; // 60 fps
+    const desiredTimeForPreview = 10;
```
lyonsno: I notice my preview buffer starts at a lower resolution than before, but it does not adjust upwards.  Is this because the lower resolution is all that my computer can handle?

src/renderer/TileRender.js
```diff
@@ -31,31 +31,17 @@ export function makeTileRender(gl) {
 
   let pixelsPerTileQuantized = pixelsPerTile;
 
-  let desiredTimePerTile = 22; // 45 fps
+  let desiredTimePerTile = 20;
```
lyonsno: Did you arrive at this number through the testing you did? If so, should we test on one more even lower spec'd device to make sure it is low enough?

src/renderer/TileRender.js
lyonsno: nice!

src/renderer/TileRender.js
lyonsno: Could you briefly explain how this works compared to before? It looks like expAvg is controlling convergence rate, similarly to the old convergenceStrength variable.  Why did you choose 0.5?

src/renderer/RenderingPipeline.js
jaxry: What is your framerate at while moving the camera?

src/renderer/TileRender.js
jaxry: The lower bound for this number is 17ms, which corresponds to 60 fps. At that point the hardware is being underworked. 20ms puts it at ~50 fps which still provided responsiveness for the devices I tested, while also being 10 FPS below the limit. That 10 FPS buffer lets the framerate fluctuate around the amount you reported, while also not hitting 60 fps, which we never want.

 In any case, testing this number on a lower spec device is important to see if what I'm saying still applies to those devices. But in any case, since the number is lower, it will run better on these devices than before this PR.

jaxry:
>Could you briefly explain how this works compared to before? It looks like expAvg is controlling convergence rate, similarly to the old convergenceStrength variable. Why did you choose 0.5?

Based on the current frame's performance, we calculate how long each pixel takes to compute, and from that derive the ideal resolution that takes the desired amount of time to sample (20 ms in this case).

The exponential averaging is not necessary, but what it does is smooth out performance differences between frames. The higher the `expAvg`, the more stable the fps becomes, but the slower it responds to changes in the scene or camera. `0.5` was a stable number for my hardware that still adjusted the framerate in about 1 second whenever I moved the camera to a more complex angle.

When your camera stays still, does your FPS do a good job hovering around 50 fps? If not, this number should be increased.

src/renderer/RenderingPipeline.js
lyonsno: 56 - 60, settling on 60 after a few seconds.  On master, which has a higher res preview buffer, framerate bounces around between 32 and 51.  Maybe the buffer on this branch just happens to start with the correct settings for my machine?

src/RayTracingRenderer.js
lyonsno: > > Could you briefly explain how this works compared to before? It looks like expAvg is controlling convergence rate, similarly to the old convergenceStrength variable. Why did you choose 0.5?
> 
> Based on the current frame's performance, we calculate how long each pixel takes to compute, and from that derive the ideal resolution that takes the desired amount of time to sample (20 ms in this case).
> 

This^ is basically the same thing that was happening before right?

> The exponential averaging is not necessary, but what it does is smooth out performance differences between frames. The higher the `expAvg`, the more stable the fps becomes, but the slower it responds to changes in the scene or camera. `0.5` was a stable number for my hardware that still adjusted the framerate in about 1 second whenever I moved the camera to a more complex angle.
> 

makes sense, how different is this than what was happening before?

> When your camera stays still, does your FPS do a good job hovering around 50 fps? If not, this number should be increased.

bounces between 47 and 53

jaxry: > This^ is basically the same thing that was happening before right?

I actually tried this current method months ago, I didn't think to use an exponential average, so the framerate varied too much with this approach. So the method before this PR steered `pixelsPerTile` to the right number using a smaller, arbitrary step that scaled with the error. That was more stable but it look longer to converge.

But now this current technique is more exact AND stable with the help of averaging.

src/renderer/RenderingPipeline.js
jaxry: The ideal FPS for the preview buffer would be something just below 60 but not actually 60. That would mean we're making max usage of the hardware, while also keeping things responsive. Maybe we can increase this number a little, but in complex scenes it will then be too high according to my testing. That's because we're currently estimating the preview size based on the running tile size. But this estimation isn't exact since it doesn't account for ray coherence across the entire scene (vs just a tile), among other issues. 

I think our best bet, in the future, would be to have a separate benchmark that just controls the preview window's size. Do you have other ideas? I'd like to tackle this during the buffer management refactor I'll be tackling in a week or two.

src/RayTracingRenderer.js
lyonsno: great, nice going!

elfrank: It would be great to avoid breaking changes when possible.
Calling this function would now cause a crash.

src/renderer/TileRender.js
elfrank: it might be good to add these magic numbers as `const` variables.

PR #57, opened 2020-01-04
Pipeline refactor

jaxry:
## Brief Description
This PR modifies and introduces new abstractions over the WebGL API with the goal of making the GL state management much more direct and easy to work with. The denoising pipeline we are introducing involves complex shaders, with many input and output buffers of different formats. This PR is an attempt at making this upcoming pipeline simple to implement without needing to worry about low-level WebGL performance gotchas and the messy API.

Here's a broad overview of the features introduces in the PR.

**RenderPass**
This class automatically handles shader compilation, uniform management, and texture allocation

- Shader Compilation: The user passes into the class constructor an object containing the shader source to compile. The object contains the source code, an array of shader chunks to import, defines to insert into the shader, as well as names of output locations for multiple render targets. The class compiles the shader with all this information, and creates a program out of the fragment shader and vertex shader passed to it.
- Uniform management: The user can set the uniform state of a program with a universal `renderPass.setUniform` method. This method works for all uniform data types, compared to WebGL's API which has separate methods for each data type. In addition, `gl.useProgram` is only called once, at draw time, to update all changed uniform values, instead of before where it was called multiple times throughout the pipeline. _This optimization increases performance of our renderer by about 10%._
- Texture allocation: Textures can be passed to the shader using the `renderer.setTexture` method. Texture binding is handled automatically, and the required textures are bound before each draw call. This allows for much easier texture manipulation on the user's end, but ends up being more work for WebGL compared to before. Beforehand, a shader reserved each texture unit throughout the lifetime of the renderer, so texture did not need to be rebound between draw calls. However, I measured no performance difference binding textures each draw call, so this abstraction is probably worth it since it allows for easier texture passing to shaders.

**Framebuffer**
This class has been redesigned from the original implementation. Beforehand it made assumptions about texture usage, and automatically created an Array Texture to be rendered to. This assumption was too limiting for what we'd like to do with the denoising pipeline. Now the class is simpler and less assuming. It simply creates a framebuffer out of textures and their corresponding render locations that are passed to it.

Lastly, some optimizations were made to the preview buffer management. Beforehand we initialized separate buffers for full-screen rendering and the reduced resolution preview rendering. Combined with all the buffers needed for reprojection, we ended up using several hundred megabytes in VRAM for buffer storage. Now, a single set of reprojection buffers are used resulting in half the amount of buffers and storage. We do this by rendering to only a smaller viewport of a full-sized buffer for preview rendering. This also makes the controls more responsive without hitches in rendering, because we don't need to keep reinitializing new preview buffers depending on system performance.

## Pull Request Guidelines


src/renderer/Framebuffer.js
```diff
@@ -1,20 +1,6 @@
...
-export function makeFramebuffer(params) {
-  const {
-    gl,
-    linearFiltering = false, // linearly filter textures
-
-    // A single render target in the form { storage: 'byte' | 'float' }
-    // Or multiple render targets passed as a RenderTargets object
-    renderTarget
-  } = params;
+export function makeFramebuffer(gl, { attachments }) {
```
jaxry: Down the line, we will add the `depthAttachment` (and possibly `stencilAttachment`) argument to the parameters to allow for depth textures as well. This will be necessary for rasterization in the hybrid rendering pipeline, but I've omitted it for now since there's no easy way to test it without the hybrid rendering code.

src/renderer/RayTracingShader.js
```diff
@@ -93,35 +78,40 @@ export function makeRayTracingShader({
       samples.restart();
     }
 
-    gl.uniform1f(uniforms.strataSize, 1.0 / strataCount);
+    renderPass.setUniform('strataSize', 1.0 / strataCount);
     nextSeed();
   }
 
+  function bindTextures() {
+    renderPass.bindTextures();
```
jaxry: Normally `renderPass` automatically binds textures when `renderPass.useProgram` is called, but we can avoid this unnecessary binding in the ray tracing shader, since the ray tracing shader is called back-to-back multiple times during tiled rendering. The old texture binding can be reused in this cased.

carlos-lopez-garces:
Thank you for the detailed description. I'll begin to review it tomorrow.

src/renderer/UniformSetter.js
```diff
@@ -0,0 +1,96 @@
...
+      v3: 0
+    };
+
+    uniforms[name] = uniform;
+  }
+
+  function setUniform(name, v0, v1, v2, v3) {
+    const uni = uniforms[name];
+
+    if (!uni) {
```
jaxry: Fail silently if uniform doesn't exist in shader. If a uniform is defined in GLSL but not actually used in some computation which renders to the screen, WebGL optimizes the shader by removing the uniform during compilation. Writing to an unused shader would error without this check. That could be useful to catch mistakes, but it's also useful to write to unused uniforms while developing shaders. Thoughts?

lyonsno: I remember we talked about a debug flag a few weeks ago, maybe errors should only be thrown in debug mode?

src/renderer/UniformSetter.js
```diff
@@ -0,0 +1,96 @@
...
+
+let typeMap;
+
+export function makeUniformSetter(gl, program) {
+  const uniformInfo = getUniforms(gl, program);
+  const uniforms = {};
+  const needsUpload = [];
+
+  for (let { name, type, location } of uniformInfo) {
+    const uniform = {
```
lyonsno: Maybe a comment here describing the intended use of v0 - v3 ?

src/renderer/UniformSetter.js
lyonsno: We could add an optional failSilently parameter?

src/renderer/UniformSetter.js
```diff
@@ -0,0 +1,96 @@
+import { getUniforms } from './glUtil';
+
+let typeMap;
+
+export function makeUniformSetter(gl, program) {
+  const uniformInfo = getUniforms(gl, program);
+  const uniforms = {};
+  const needsUpload = [];
+
+  for (let { name, type, location } of uniformInfo) {
```
lyonsno: Also, why isn't this rolled up into the getUniforms function from glUtils?

src/renderer/glsl/chunks/core.glsl
```diff
@@ -0,0 +1,93 @@
+export default `
```
lyonsno: Nice moving this to its own file

src/renderer/UniformSetter.js
jaxry: I just added a new commit trying a new idea. The console warns (once) if a uniform doesn't exist, but the error is handled and the program keeps running. This way we can have unused uniforms in the shader during development, but we're still aware of any uniforms that are aren't being set due to a mismatch in names or whatever. This way we don't need to deal with the complexity of a debug flag. How does this sound?

lyonsno: perfect.

src/renderer/UniformSetter.js
jaxry: My intention was that glUtil's `getUniforms` is concerned with fetching an array of uniform descriptions from the shader, while the object here is concerned with storing variables to write to the shader. That way we have a separation of concerns. What are you thinking we could do?

src/renderer/glsl/rayTrace.frag
```diff
@@ -8,242 +10,144 @@ import sampleMaterial from './chunks/sampleMaterial.glsl';
...
+  random,
+  envmap,
+  bsdf,
+  sample,
+  sampleMaterial,
+  sampleGlass,
+  sampleShadowCatcher,
+],
+outputs: ['light', 'position'],
+source: (defines) => `
```
lyonsno: Loving this so much. You are making me very happy @jaxry :)

src/renderer/RenderingPipeline.js
```diff
@@ -19,101 +20,124 @@ export function makeRenderingPipeline({
...
+          [rayTracingShader.outputs.light]: floatTex(),
+          [rayTracingShader.outputs.position]: floatTex(),
+        }
+      });
 
-  // how many samples to render with uniform noise before switching to stratified noise
-  const numUniformSamples = 6;
+    const makeReprojectBuffer = () => makeFramebuffer(gl, {
+        attachments: { 0: floatTex() }
+      });
```
jaxry: Are these parameters confusing? For `attachments`, the property name is a number specifying the location index, and the value is a texture. It's a brief way to write it but should we be more verbose? `attachments` could be in the form

```javascript
const rayTracingAttachments = [
  {
    location: rayTracingShader.outputs.light,
    texture: floatTex()
  },
  {
    location: rayTracingShader.outputs.position,
    texture: floatTex()
  }
];

makeFrameBuffer(gl, { attachments: rayTracingAttachments })
```
and

```javascript
const reprojectionAttachments = [
  {
    location: 0,
    texture: floatTex()
  }
];

makeFrameBuffer(gl, { attachments: reprojectionAttachments })

File: src/renderer/RenderingPipeline.js [hunk 9]
```diff
@@ -253,55 +313,45 @@ export function makeRenderingPipeline({
...
-    reprojectShader.draw(hdrBuffer.texture, historyBuffer.texture);
+    gl.viewport(0, 0, screenWidth, screenHeight);
+    reprojectShader.draw({
+      blendAmount: 1.0,
+      light: hdrBuffer.attachments[rayTracingShader.outputs.light],
+      position: hdrBuffer.attachments[rayTracingShader.outputs.position],
+      previousLight: reprojectBackBuffer.attachments[0],
+      previousPosition: hdrBackBuffer.attachments[rayTracingShader.outputs.position],
+      textureScale: fullscreenScale,
+      previousTextureScale: fullscreenScale
```
jaxry: This section is uglier than before, but it's because we're manually specifying samplers and attachment points now. Before we assumed everything was packed neatly in an Array Texture, which made this syntax short and sweet. But with hybrid rendering, the position texture will come from the raster pass, and the light texture will come from the ray tracing pass. This new granularity will be useful then.

src/renderer/glsl/reproject.frag
```diff
@@ -1,102 +1,102 @@
...
+source: `
+  in vec2 vCoord;
+
+  uniform mediump sampler2D light;
+  uniform mediump sampler2D position;
+  uniform vec2 textureScale;
+
+  uniform mediump sampler2D previousLight;
+  uniform mediump sampler2D previousPosition;
+  uniform vec2 previousTextureScale;
```
lyonsno: way way more readable

src/renderer/UniformSetter.js
lyonsno: Do you foresee other uses for getUniforms?

src/renderer/glsl/reproject.frag
```diff
@@ -1,102 +1,102 @@
...
-  float sum;
+    vec4 history;
+    float sum;
 
-  // bilinear sampling, rejecting samples that don't have a matching mesh id
-  for (int i = 0; i < 4; i++) {
-    float histMeshId = texelFetch(historyBuffer, ivec3(texel[i], historyBuffer_position), 0).w;
+    // bilinear sampling, rejecting samples that don't have a matching mesh id
+    for (int i = 0; i < 4; i++) {
+      float histMeshId = texelFetch(previousPosition, texel[i], 0).w;
```
lyonsno: This is also a huge improvement

src/renderer/glsl/reproject.frag
```diff
@@ -1,102 +1,102 @@
...
-    history.xyz *= MAX_SAMPLES / history.w;
-    history.w = MAX_SAMPLES;
-  }
+    if (history.w > MAX_SAMPLES) {
+      history.xyz *= MAX_SAMPLES / history.w;
+      history.w = MAX_SAMPLES;
+    }
 
-  out_light = blendAmount * history + lightTex;
-  out_position = positionTex;
```
lyonsno: I forget, does removing the need for this line boost performance?

src/renderer/glsl/toneMap.frag
```diff
@@ -1,64 +1,59 @@
...
-}
-
-`;
+export default {
+includes: [textureLinear],
+outputs: ['color'],
+source: `
+  in vec2 vCoord;
+
+  uniform mediump sampler2D hdrBuffer;
```
lyonsno: We're calling this 'light' in a lot of other places, is that mostly to differentiate it from 'position' or should we continue that convention here?

src/renderer/RenderingPipeline.js
```diff
@@ -1,13 +1,14 @@
...
 import { makeTileRender } from './TileRender';
-import { makeTextureAllocator } from './TextureAllocator';
+import { makeTexture } from './Texture';
 import { makeReprojectShader } from './ReprojectShader';
 import noiseBase64 from './texture/noise';
-import { PerspectiveCamera } from 'three';
+import { clamp } from './util';
+import { PerspectiveCamera, Vector2 } from 'three';
 
 // Important TODO: Refactor this file to get rid of duplicate and confusing code
```
lyonsno: would it be possible to sum up which elements are confusing in one line or less?

src/renderer/RenderingPipeline.js
```diff
@@ -19,101 +20,124 @@ export function makeRenderingPipeline({
...
+
+  // how many partitions of stratified noise should be created
+  // higher number results in faster convergence over time, but with lower quality initial samples
+  const strataCount = 6;
 
-  const textureAllocator = makeTextureAllocator(gl);
+  const fullscreenQuad = makeFullscreenQuad(gl);
 
-  const rayTracingShader = makeRayTracingShader({bounces, fullscreenQuad, gl, optionalExtensions, scene, textureAllocator});
+  const rayTracingShader = makeRayTracingShader(gl, { bounces, fullscreenQuad, optionalExtensions, scene });
```
lyonsno: I like the separation of gl from the other params

src/renderer/RenderingPipeline.js
```diff
@@ -19,101 +20,124 @@ export function makeRenderingPipeline({
...
 
-  const fullscreenQuad = makeFullscreenQuad(gl);
+  // how many samples to render with uniform noise before switching to stratified noise
+  const numUniformSamples = 6;
+
+  // how many partitions of stratified noise should be created
+  // higher number results in faster convergence over time, but with lower quality initial samples
+  const strataCount = 6;
 
-  const textureAllocator = makeTextureAllocator(gl);
```
lyonsno: For my own understanding:
This textureallocator keeps track of texture units globally to ensure that textures are not bound to overlapping units. This could potentially exceed the maxTextureUnits where it wouldn't have to, in a situation where even though each pass used fewer than max units, the total number of textures used by all passes would exceed max.   The new method where texture units are kept track of individually per render pass could potentially support an infinite number of passes as long as each one did not exceed maxTextureUnits. Is this correct?

src/renderer/RenderingPipeline.js
lyonsno: This is verbosity that's easy to collapse. I think the verbose version is certainly clearer, so I think it's worth the tradeoff to be more verbose here.

src/renderer/RenderingPipeline.js
```diff
@@ -19,101 +20,124 @@ export function makeRenderingPipeline({
...
 
-  // how many partitions of stratified noise should be created
-  // higher number results in faster convergence over time, but with lower quality initial samples
-  const strataCount = 6;
+    hdrBuffer = makeHdrBuffer();
+    hdrBackBuffer = makeHdrBuffer();
 
-  let sampleCount = 1;
+    reprojectBuffer = makeReprojectBuffer();
+    reprojectBackBuffer = makeReprojectBuffer();
```
lyonsno: Is this a stylistic choice or is there another reason for it?

src/renderer/RenderPass.js
```diff
@@ -0,0 +1,170 @@
+import { compileShader, createProgram } from './glUtil';
+import { makeUniformSetter } from './UniformSetter';
+
+export function makeRenderPass(gl, params) {
```
lyonsno: @jaxry what do you think about calling this ShaderPass instead of RenderPass? It seems strange that each shader object owns its own renderpass, I usually think of a render pass as using a shader, not the other way around.
I could totally be missing something here though.

src/renderer/RenderingPipeline.js
```diff
@@ -19,101 +20,124 @@ export function makeRenderingPipeline({
...
+  }
+
+  // Shaders will read from the back buffer and draw to the front buffer
+  // Buffers are swapped after every render
+  function swapBuffers() {
+    swapHdrBuffer();
+    swapReprojectBuffer();
+  }
+
+  function setSize(w, h) {
```
lyonsno: This is going to be used only for changes in primary window size, not for controlling preview resolution, correct?

src/renderer/RenderPass.js
```diff
@@ -0,0 +1,170 @@
...
+
+  const vertexCompiled = vertex instanceof WebGLShader ? vertex : makeVertexShader(gl, params);
+
+  const fragmentCompiled = fragment instanceof WebGLShader ? fragment : makeFragmentShader(gl, params);
+
+  const program = createProgram(gl, vertexCompiled, fragmentCompiled);
+
+  return {
+    ...makeRenderPassFromProgram(gl, program),
+    outputs: getOutputLocations(fragment.outputs)
```
lyonsno: what do you think of changing `outputs` to `outputLocations`?

src/renderer/RenderingPipeline.js
lyonsno: It's messy, but it's not confusing. I think it's ok, especially for now, I actually found it a little helpful when looking at the code for the first time.

src/renderer/RenderingPipeline.js
```diff
@@ -253,55 +313,45 @@ export function makeRenderingPipeline({
...
-    tileRender.setSize(w, h);
-    hdrBuffer.setSize(w, h);
-    reprojectBuffer.setSize(w, h);
-    initFirstSample();
+    swapReprojectBuffer();
   }
 
   return {
-    drawTile,
+    draw,
```
lyonsno: I'm guessing this change was because tile rendering is always going to be the default mode. Even so, I'm not completely convinced, I don't necessarily think that being explicit is bad here.

src/renderer/UniformSetter.js
jaxry: I can't think of any uses we'd actually be using in near future.

src/renderer/glsl/reproject.frag
jaxry: I didn't measure a difference

src/renderer/glsl/toneMap.frag
jaxry: I like the idea of continuing that convention here to make everything consistent.

src/renderer/RenderingPipeline.js
jaxry: I think that comment was left over from much earlier... I'll remove it all together.

src/renderer/RenderingPipeline.js
jaxry: Correct. The downside is that there's a potential performance cost of binding textures each render pass. I was very careful with measurements and I did not see a difference. But I'll be keeping an eye on this approach down the road to make sure it doesn't ever slow anything down.

src/renderer/RenderingPipeline.js
jaxry: Could you expand your question? I'm not sure what you're referring to.

src/renderer/UniformSetter.js
lyonsno: Ok neither could I.  In light of that it seemed a little redundant to do it this way, but separation of concerns is a pretty good argument. I'm satisfied.

src/renderer/glsl/reproject.frag
lyonsno: But now we don't have to bother to explicitly write it to preserve the position texture, right?

src/renderer/RenderPass.js
jaxry: I'm a little confused which one makes more sense. It was originally ShaderPass, but @carlos-lopez-garces mentioned [this tutorial](http://cwyman.org/code/dxrTutors/tutors/Tutor2/tutorial02.md.html) which uses the term RenderPass in a similar way to what we're doing here (a GLProgram that links shaders and uniforms). I think I'm partial to RenderPass, because technically that object encapsulates a program linking two shaders, and not just a single fragment or vertex shader.

But I see your concern. Multiple render passes could use the same shader in a hypothetical scenario, so it's confusing what we should do.

src/renderer/RenderingPipeline.js
jaxry: Correct. Preview resolution is just controlled via the `textureScale` uniform now.

src/renderer/RenderPass.js
jaxry: I like that. Would `outputLocs` make just as much sense, or should we go for the longer form?

src/renderer/RenderingPipeline.js
jaxry: I agree being explicit is good, but moving forward we might be moving away from tiles altogether once we have hybrid rendering and spatial denoising in place. My intent was that `pipeline.draw` is supposed to be a black box that the main RayTracingRenderer calls to generate an image. Whatever happens in there is up to the internal engine. That way the method makes semantic sense regardless of the specifics of the pipeline.

src/renderer/glsl/reproject.frag
jaxry: Yeah. The reproject buffer just has to worry about reprojecting light.

src/renderer/RenderingPipeline.js
lyonsno: packing the buffer creation with params, etc. into the `makeReprojectBuffer` type functions. Except I just realized that it's to prevent duplicated code, so nvm 🤪

src/renderer/RenderingPipeline.js
lyonsno: Ok, I see your point.

src/renderer/RenderPass.js
lyonsno: `outputLocs` is fine, I've seen that convention plenty of times.

src/renderer/RenderingPipeline.js
```diff
@@ -19,101 +18,124 @@ export function makeRenderingPipeline({
...
 
-  const rayTracingShader = makeRayTracingShader({bounces, fullscreenQuad, gl, optionalExtensions, scene, textureAllocator});
+  const rayTracingShader = makeRayTracingShader(gl, { bounces, fullscreenQuad, optionalExtensions, scene });
 
-  const reprojectShader = makeReprojectShader({ fullscreenQuad, gl, maxReprojectedSamples, textureAllocator });
+  const reprojectShader = makeReprojectShader(gl, { fullscreenQuad, maxReprojectedSamples });
 
-  const toneMapShader = makeToneMapShader({
-    fullscreenQuad, gl, optionalExtensions, textureAllocator, toneMappingParams
+  const toneMapShader = makeToneMapShader(gl, {
```
jaxry: Thoughts on renaming all of these to `rayTracingPass`, `reprojectPass` and `toneMapPass`?

src/renderer/RenderingPipeline.js
jaxry: Do you still think the tradeoff is worth it with the rename to `outputLocs`?

src/renderer/RenderPass.js
lyonsno: > I'm a little confused which one makes more sense. It was originally ShaderPass, but @carlos-lopez-garces mentioned [this tutorial](http://cwyman.org/code/dxrTutors/tutors/Tutor2/tutorial02.md.html) which uses the term RenderPass in a similar way to what we're doing here (a GLProgram that links shaders and uniforms). I think I'm partial to RenderPass, because technically that object encapsulates a program linking two shaders, and not just a single fragment or vertex shader.
> 

This is a good point. Maybe a solution would be to rename each of our .*Shader.js files to .*RenderPass or even .*Pass?
e.g. `ToneMappingShader.js` => `ToneMappingRenderPass.js` or `ToneMappingPass.js`

> Multiple render passes could use the same shader in a hypothetical scenario

I don't understand exactly what you mean here.

src/renderer/RenderingPipeline.js
lyonsno: Haha I think we agree on this. I think it's an improvement to readability from a high level as well. Nice!

src/renderer/RenderingPipeline.js
lyonsno: How do you plan on doing that?

src/renderer/RenderPass.js
jaxry: Okay I think we should make this change then. I like it.

>Multiple render passes could use the same shader in a hypothetical scenario

Nevermind, we're on the same page after all :)

src/renderer/RenderingPipeline.js
jaxry: Once we significantly change the pipeline with hybrid rendering and whatnot, I'll just hard code static texture allocation with global variables inside RenderPass and measure performance then. If it turns out it's faster doing it like that, we can think about then how to implement static textures in a cleaner way than it was previously. How about that?

lyonsno: > Once we significantly change the pipeline with hybrid rendering and whatnot, I'll just hack in static texture allocation with global variables inside RenderPass and measure performance then. If it turns out to be faster, we can decide then how implement static textures in a cleaner way than it was previous to this PR. How about that?
> 
> Edit: I played around with smart static texture binding while making this PR and it's definitely possible to persist texture bindings intelligently without needing to change the API. But it's hard to implement because there are a lot of edge cases to consider (namely, what to do if you run out of texture units?). I figured it's not worth the headache if there's no performance improvement

I agree completely. Nice job on this.

lyonsno:
I just did some comparisons with the QA branch, and I noticed a couple of discrepancies.

Most significantly in the behavior of reprojection.  The changes from this PR cause the reprojected preview buffer to jitter dramatically while it blends together with the high res buffer. This is not a great explanation but I have to head out for a bit so hopefully you can understand what I mean.

To reproduce, try orbiting the camera quickly in the test-scene, and then once it stops moving pay attention to the bright highlight speckles on the glass spheres. You should notice these specks jittering around in the background as they fade into the higher res image as it converges.

jaxry:
Oops, very good catch. This bug just happened recently so I'll figure out the cause.

lyonsno:
Nice, the jitter seems to be fixed!

lyonsno:
Now I'm driving myself nuts trying to figure out if there are any other differences in the reprojected preview buffer during camera movement. I thought there was at first but I have no idea how to compare in a more rigorous fashion. I'm starting to think it was just in my head...

Everything else seems good to go though!

src/renderer/UniformSetter.js
```diff
@@ -0,0 +1,105 @@
+import { getUniforms } from './glUtil';
+
+let typeMap;
+
+export function makeUniformSetter(gl, program) {
```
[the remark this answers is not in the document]
jaxry: UniformSetter is designed to be used by anything that has a GLProgram, which wants to set uniforms on it. For now, RenderPass is the only class which does so, but their functionality is separate and isn't strictly tied together. I can't think of any useful examples for us right now, but there may be a reason why we'd want GLPrograms outside of RenderPass. I like the idea of having that possibility of using UniformSetter in different ways, and not making assumptions about the class's use by imposing some import restriction on it.

> contributors might decide to use it and forgetful reviewers may not prevent it

The only way this would be bad is if the contributer is reimplementing RenderPass without realizing its existence, but it's one of the core library classes, so how would that happen? In any other novel case, the use of UniformSetter would be interesting.

I might not be understanding the full picture of what you're talking about.

src/renderer/RenderPass.js
```diff
@@ -0,0 +1,170 @@
+import { compileShader, createProgram } from './glUtil';
+import { makeUniformSetter } from './UniformSetter';
+
+export function makeRenderPass(gl, params) {
+  const {
+    defines,
```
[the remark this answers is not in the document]
jaxry: That's true. I left it there so that anyone opening the RenderPass file to look at the parameter structure can see exactly what the constructor expects. Even though it isn't used in the destructured form right there, the `params` object is passed to child functions which do use `defines`. Omitting it might obscure the fact that it's a valid argument to supply `makeRenderPass`. What are the advantages of omitting it?

src/renderer/RenderPass.js
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  str += addIncludes(shader.includes, defines);
+
+  if (typeof shader.source === 'function') {
+    str += shader.source(defines);
+  } else if (typeof shader.source === 'string') {
+    str += shader.source;
+  } else {
+    console.error('Provide a shader source string');
```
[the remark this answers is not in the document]
jaxry: Whenever I use the term "user" in this PR, I'm strictly referring to any class which uses the class in question. Regardless of whether the user of the _library_ uses the class. I agree the custom error messages aren't that helpful. More info in my other comment.

jaxry:
>Do you have any ideas how to achieve more certainty here?

Not any immediate solutions, and this will be a problem down the line. I think we need a way to generate deterministic random numbers so that we can compare exact frames between branches. For now the visual component is in such constant state of flux I think we're _okay_ with putting it off, but once spatial filtering and the like is in, it will become more important. Thoughts?

If everything else looks good, I'd like to merge so that we can start porting over hybrid rendering asap. I know @carlos-lopez-garces is still reviewing, and I'm sure there are still problems with the PR. If it's possible I'd like to make those changes in a separate PR so that hybrid rendering isn't blocked. Or is there another way we can do it?

src/renderer/UniformSetter.js
```diff
@@ -0,0 +1,105 @@
...
+function initTypeMap(gl) {
+  return {
+    [gl.FLOAT]: glName(1, 'f'),
+    [gl.FLOAT_VEC2]: glName(2, 'f'),
+    [gl.FLOAT_VEC3]: glName(3, 'f'),
+    [gl.FLOAT_VEC4]: glName(4, 'f'),
+    [gl.INT]: glName(1, 'i'),
+    [gl.INT_VEC2]: glName(2, 'i'),
+    [gl.INT_VEC2]: glName(3, 'i'),
+    [gl.INT_VEC2]: glName(4, 'i'),
```
carlos-lopez-garces: `INT_VEC2` appears 3 times as map key. A typo, probably. I think this is what we want:

```js
[gl.INT_VEC2]: glName(2, 'i'),
[gl.INT_VEC3]: glName(3, 'i'),
[gl.INT_VEC4]: glName(4, 'i'),
```

src/renderer/RenderPass.js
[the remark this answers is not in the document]
carlos-lopez-garces: > What are the advantages of omitting it?

None, really. Linters warn you about unused variables, if you choose to be warned (like Machete's). I guess that's why I left the comment.

Your reasons for leaving it there justify it.

src/renderer/UniformSetter.js
```diff
@@ -0,0 +1,105 @@
...
+    values: `uniform${numComponents}${type}`,
+    array: `uniform${numComponents}${type}v`
+  };
+}
+
+function glNameMatrix(rows, columns) {
+  return {
+    matrix: rows === columns ?
+      `uniformMatrix${rows}fv` :
+      `uniformMatrix${rows}x${columns}fv`
```
carlos-lopez-garces: The fact that `typeMap` doesn't include entries for non-square matrix types and the fact that `glNameMatrix()` here does consider and handle them make me wonder:

a) Is `typeMap` missing those types by accident (`FLOAT_MAT2x3`, `FLOAT_MAT3x4`, etc)? From comments you've made, I understand that we want `UniformSetter` (and the rest of the new GL abstractions) to be able to handle any shader code it may get passed (and not just the handful of vertex and fragment shaders that exist at the moment). If _we_ added a new fragment shader that needed to use one of the uniform types that are not registered in `typeMap` (like the non-square matrices I mentioned, or even `gl.UNSIGNED_INT` and `gl. SAMPLER_2D_SHADOW`), we'd need to register it as well. If that new fragment shader didn't come from us, but from the user who's using the renderer as a library (my understanding is that they may use these GL abstractions directly), would he be able to use any type of uniform without having to register it first?

or

b) Is the "else" branch of `glNameMatrix()`'s conditional (i.e. `uniformMatrix${rows}x${columns}fv`) dead code?

jaxry: >Is typeMap missing those types by accident (FLOAT_MAT2x3, FLOAT_MAT3x4, etc)?

No, they're omitted on purpose.

>If we added a new fragment shader that needed to use one of the uniform types that are not registered in typeMap, we'd need to register it as well. 

Correct. As `glNameMatrix` implies, adding support should just involve adding the corresponding types to the `typeMap`. A complete UniformSetter implementation should include those, but since we're not using any Mat2x3 (etc.) uniforms in the renderer, I wasn't able to test whether those type maps actually work as intended. I would have to write a test shader that uses these uniforms to verify everything works, but that felt like more trouble than it was worth, for now. Do you think it's better to leave the implementation incomplete (and just add a comment noting that) or should we add the remaining types but keep them untested (they will probably work but I'm not sure)? Testing it seems like too much work for what we get right now, since I don't forsee us ever using those types...

src/renderer/UniformSetter.js
jaxry: Yikes. Copy-pasting is dangerous

carlos-lopez-garces:
> If everything else looks good, I'd like to merge so that we can start porting over hybrid rendering asap. I know @carlos-lopez-garces is still reviewing, and I'm sure there are still problems with the PR. If it's possible I'd like to make those changes in a separate PR so that hybrid rendering isn't blocked. Or is there another way we can do it?

@jaxry The sad truth is that if we merge it now, I'll probably switch my attention to something else and never come back (that's what happened with me with the reprojection PR). You can always keep this one open and rebase the hybrid rendering branch on top of this one, and set that PR's base branch to this one as well.

But I'm not very familiar with the renderer's release plan and I wouldn't want to be the one that holds it up. If the rebase option I offered doesn't work, and you need to merge this, by all means do.

jaxry:
>You can always keep this one open and rebase the hybrid rendering branch on top of this one, and set that PR's base branch to this one as well.

That works perfectly well for dev purposes as long as we consider two things I can think of.

1. If I start working on hybrid rendering now, and there's something fundamentally wrong with the API here, updating both branches is trouble. Does the API from an external perspective look solid to you?

2. This branch comes with performance improvements (10% faster rendering due to less program switches, and no preview buffer resize hitches), so it might be nice to see in Machete and Design Pro.

carlos-lopez-garces:
> If I start working on hybrid rendering now, and there's something fundamentally wrong with the API here, updating both branches is trouble.

That's what I've been doing with a series of refactors I'm working on over at Machete, and yes, it's a pain.

> Does the API from an external perspective look solid to you?

The truth is I've only looked at `RenderPass` and `UniformSetter` and I still don't have a complete picture of the whole API. But I can prioritize that next, look at the API instead of dwelling on the details to expedite the review. 👍

carlos-lopez-garces:
As much as I want to finish my review, I need to focus on my project completely for the next few days. Feel free to merge with @lyonsno's sign-off.

src/renderer/RenderPass.js
```diff
@@ -0,0 +1,170 @@
...
+
+  return compileShader(gl, type, str);
+}
+
+function addDefines(defines) {
+  let str = '';
+
+  if (typeof defines !== 'object') {
+    return str;
+  }
```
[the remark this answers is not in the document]
carlos-lopez-garces: > However, keeping our code lean and not doing unnecessary input validation is important too. Which is preferable in this case?

I think both things are important in this case and also in general. While we can rely on our intuition to keep the code lean, we can't just do that for judging input validation, I think we need to think about it seriously in the context of the entire codebase and agree on a strategy and conventions. I think this project is mature enough to have that, and at the same time not too large to be too late for that.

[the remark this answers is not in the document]
carlos-lopez-garces: > we should be using Typescript or something similar 

I like that.

src/renderer/RenderPass.js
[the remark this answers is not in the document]
carlos-lopez-garces: Got it. It might take a while until I can picture exactly how users of the library will use these APIs.

src/renderer/UniformSetter.js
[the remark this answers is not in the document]
carlos-lopez-garces: > I might not be understanding the full picture of what you're talking about.

I was under the impression that `RenderPass` was the gateway to all or most GL operations; that if anybody (the library or the client) wanted to do GL operations, they'd do so through `RenderPass`'s interface (sound silly when I write it, given that `RenderPass` appears to be so much more than that). Under that impression, it seemed reasonable to try to enforce that decision with some language mechanism.

But you've explained that that's not the case.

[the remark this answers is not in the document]
jaxry: >I was under the impression that RenderPass was the gateway to all or most GL operations

It is, in a way. RenderPass adds abstractions which make shader management simple, but it's an abstraction because it automatically handles texture management, uniform management, etc. in an assuming way. The assumptions come from how we've been using shaders up until this point, but if we get to a step where we need to implement some wacky memory management in the hybrid pipeline, RenderPass might not be sufficient. In that case I'm thinking UniformSetter might still be useful for some other GL abstraction that otherwise diverges from RenderPass. It's all hypothetical for now, but that's my reasoning for keeping UniformSetter separate from RenderPass.

jaxry:
@lyonsno if everything looks good to you, could you approve the PR?

PR #58, opened 2020-01-13
Tweak reprojection amount & decrease bounces to 2

jaxry:
## Brief Description
This PR decreases the amount of reprojection each frame, decreasing the amount of ghosting and low resolution pixelation, at the cost of extra noise. However, the slight increase in noise looks better than the amount of ghosting present beforehand.

In addition, the default `bounces` is now set to 2. This value drastically improves performance, and is sufficient for the majority of scenes, without noticeable loss of realism. Most denoising pipelines in research papers use 2 bounces, so it makes sense to use this as a default moving forward.

## Pull Request Guidelines

### DIFF
--- a/scenes/webgl-comparison/main.js
+++ b/scenes/webgl-comparison/main.js
@@ -154,6 +154,7 @@ function initRenderer(renderer) {
   renderer.toneMapping = THREE.ACESFilmicToneMapping;
   renderer.toneMappingExposure = 1.5;
   renderer.renderWhenOffFocus = false;
+  renderer.bounces = 3;
 }
 
 function unloadRenderer(renderer) {
--- a/src/renderer/RenderingPipeline.js
+++ b/src/renderer/RenderingPipeline.js
@@ -21,7 +21,7 @@ export function makeRenderingPipeline({
   const maxReprojectedSamples = 20;
 
   // how many samples to render with uniform noise before switching to stratified noise
-  const numUniformSamples = 3;
+  const numUniformSamples = 4;
 
   // how many partitions of stratified noise should be created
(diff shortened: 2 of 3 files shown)


scenes/webgl-comparison/main.js
```diff
@@ -154,6 +154,7 @@ function initRenderer(renderer) {
   renderer.toneMapping = THREE.ACESFilmicToneMapping;
   renderer.toneMappingExposure = 1.5;
   renderer.renderWhenOffFocus = false;
+  renderer.bounces = 3;
```
jaxry: The readme scene has enough indirect illumination that it needs 3 bounces.

src/renderer/RenderingPipeline.js
```diff
@@ -18,8 +18,7 @@ export function makeRenderingPipeline({
     bounces, // number of global illumination bounces
   }) {
 
-  const reprojectDecay = 0.975;
```
jaxry: Remove this decay value altogether and use `1` in its place. The averaging of samples in the reprojection shader automatically adds decay to old samples, so we don't need to add more.

src/renderer/RenderingPipeline.js
```diff
@@ -18,8 +18,7 @@ export function makeRenderingPipeline({
...
-  const maxReprojectedSamples = Math.round(reprojectDecay / (1 - reprojectDecay));
+  const maxReprojectedSamples = 25;
```
jaxry: The previous value calculated out to be 39. The lower value of 25 provides the best balance of noise vs ghosting in my opinion for my laptop, tested on a HOVER scene as well as the demo scenes.

jaxry:
Still need an approval to merge.

PR #62, opened 2020-01-20
Hybrid rendering

jaxry:
## Brief Description
Work in progress.

This PR adds hybrid rendering to the renderer. This means we first rasterize the scene to several g-buffers (position, normal, albedo, etc). From there we start the ray tracing step from the g-buffers instead of the camera. 

Advantages
* Skip the first intersection. Provides a 25-50% performance boost in practice.
* G-buffers can always be full resolution. This lets us render everything at native resolution, all the time, while only upscaling global illumination. This promises massive visual quality improvements. 

## Pull Request Guidelines


jaxry:
How can I rebase this branch so that it doesn't include the commits of the pipeline refactor https://github.com/hoverinc/ray-tracing-renderer/pull/57?

src/renderer/MaterialBuffer.js
```diff
@@ -0,0 +1,175 @@
+import { ThinMaterial, ThickMaterial, ShadowCatcherMaterial } from '../constants';
```
jaxry: Move all the material processing in RayTracePass.js and UploadBuffer.js to a single, independent file that processes materials and exports a material uniform buffer. This uniform buffer is used by RayTracePass and GBufferPass

src/renderer/Texture.js
```diff
@@ -24,10 +24,10 @@ export function makeTexture(gl, params) {
...
     // sampling properties
     gammaCorrection = false,
-    wrapS = gl.REPEAT,
-    wrapT = gl.REPEAT,
-    minFilter = gl.LINEAR,
-    magFilter = gl.LINEAR,
+    wrapS = gl.CLAMP_TO_EDGE,
+    wrapT = gl.CLAMP_TO_EDGE,
+    minFilter = gl.NEAREST,
+    magFilter = gl.NEAREST,
```
jaxry: Use the WebGL defaults instead of our own overridden defaults.

src/renderer/UniformBuffer.js
```diff
@@ -0,0 +1,92 @@
+export function makeUniformBuffer(gl, program, blockName) {
```
jaxry: Move from glUtil.js to its own file.

src/renderer/glsl/chunks/materialBuffer.glsl
```diff
@@ -0,0 +1,100 @@
+export default `
+
+uniform Materials {
```
jaxry: Abstract material fetching so that RayTrace.frag and GBuffer.frag both use the same interface.

src/renderer/glsl/chunks/sampleMaterial.glsl
```diff
@@ -97,6 +97,10 @@ void sampleMaterial(SurfaceInteraction si, int bounce, inout Path path) {
 
   // Get new path direction
 
+  if (lastBounce) {
```
jaxry: Possible optimization. On some scenes this provides a 5% boost.

src/renderer/glsl/reproject.frag
```diff
@@ -5,33 +5,78 @@ source: `
...
 
   uniform mediump sampler2D previousLight;
   uniform mediump sampler2D previousPosition;
-  uniform vec2 previousTextureScale;
 
   uniform mat4 historyCamera;
   uniform float blendAmount;
   uniform vec2 jitter;
 
+  vec4 getUpscaledLight(vec2 coord) {
```
jaxry: This function is messy. Cleanup with similar code in main(), and replace with cubic filtering...?

jaxry:
Ready for review!

src/renderer/UniformSetter.js
```diff
@@ -28,10 +30,10 @@ export function makeUniformSetter(gl, program) {
...
 
     if (!uni) {
-      if (!failedUnis.has(name)) {
-        console.warn(`Uniform "${name}" does not exist in shader`);
-        failedUnis.add(name);
-      }
+      // if (!failedUnis.has(name)) {
+      //   console.warn(`Uniform "${name}" does not exist in shader`);
+      //   failedUnis.add(name);
+      // }
```
jaxry: We sometimes set a uniform for an optional feature that doesn't exist for a particular scene (e.g. material normal maps). This behavior is intended since it simplifies the init logic. Emitting warning messages doesn't make sense in this case so I've commented this out until we figure out what we really want to do here.

src/renderer/glsl/toneMap.frag
```diff
@@ -37,15 +38,62 @@ source: `
...
+    }
+
+    if (sum > 0.0) {
+      upscaledLight /= sum;
+    } else {
+      upscaledLight = texture(light, lightScale * coord);
+    }
+
+    return upscaledLight;
+  }
```
lyonsno: Anything specific you'd like me to focus on for this function?

src/renderer/mergeMeshesToGeometry.js
```diff
@@ -40,51 +40,51 @@ export function mergeMeshesToGeometry(meshes) {
...
-
-  const bg = new BufferGeometry();
-  bg.addAttribute('position', position);
-  bg.addAttribute('normal', normal);
-  bg.addAttribute('uv', uv);
-  bg.setIndex(index);
+  const positionAttrib = new BufferAttribute(new Float32Array(3 * vertexCount), 3, false);
+  const normalAttrib = new BufferAttribute(new Float32Array(3 * vertexCount), 3, false);
+  const uvAttrib = new BufferAttribute(new Float32Array(2 * vertexCount), 2, false);
+  const materialMeshIndexAttrib = new BufferAttribute(new Float32Array(2 * vertexCount), 2, false);
```
lyonsno: Could you give some context for this change?

src/renderer/glsl/gBuffer.frag
```diff
@@ -0,0 +1,57 @@
...
+
+  vec3 faceNormals(vec3 pos) {
+    vec3 fdx = dFdx(pos);
+    vec3 fdy = dFdy(pos);
+    return cross(fdx, fdy);
+  }
+
+  void main() {
+    int materialIndex = int(EPS + vMaterialMeshIndex.x);
+    float meshIndex = floor(EPS + vMaterialMeshIndex.y);
```
lyonsno: why add an EPS and why floor vs int?

src/renderer/glsl/rayTrace.frag
```diff
@@ -115,20 +123,13 @@ source: (defines) => `
     Ray cam;
     initRay(cam, origin, direction);
 
-    SurfaceInteraction si;
-
-    vec4 liAndAlpha = integrator(cam, si);
-
-    if (dot(si.position, si.position) == 0.0) {
-      si.position = origin + direction * RAY_MAX_DISTANCE;
-    }
```
lyonsno: Why remove?

src/renderer/glsl/rayTrace.frag
```diff
@@ -115,20 +123,13 @@ source: (defines) => `
     Ray cam;
     initRay(cam, origin, direction);
 
-    SurfaceInteraction si;
```
lyonsno: I'm guessing we can move this into the integrator function because we no longer need line 131, correct?

src/renderer/glsl/reproject.frag
```diff
@@ -21,17 +24,20 @@ source: `
...
-    vec4 positionTex = texture(position, scaledCoord);
-    vec4 lightTex = texture(light, scaledCoord);
+    vec4 currentLight = texture(light, lightScale * vCoord);
 
-    vec3 currentPosition = positionTex.xyz;
-    float currentMeshId = positionTex.w;
+    if (currentMeshId == 0.0) {
+      out_light = currentLight;
+      return;
+    }
```
lyonsno: Is this related to hybrid rendering or is it a generic improvement to the reprojection algorithm?

src/renderer/glsl/gBuffer.frag
```diff
@@ -0,0 +1,57 @@
...
+      vec2 duv1 = dFdx(vUv);
+      vec2 duv2 = dFdy(vUv);
+      normal = getMatNormal(materialIndex, uv, normal, dp1, dp2, duv1, duv2);
+    #endif
+
+    out_position = vec4(vPosition, meshIndex);
+    out_normal = vec4(normal, materialType);
+    out_faceNormal = vec4(faceNormal, 0);
+    out_color = vec4(color, 0);
+    out_matProps = vec4(roughness, metalness, 0, 0);
```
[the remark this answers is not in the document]
jaxry: I'm not sure. We could pack roughness/metalness into the other textures, but I tried that and I didn't notice a performance hit. So I kept them separate since we can use the empty channels to separate out the envmap next.

src/renderer/glsl/toneMap.frag
jaxry: Not in particular, unless there's something sticking out to you. I'll add a comment describing what this function does.

src/renderer/mergeMeshesToGeometry.js
jaxry: Before, the material index was just stored in the BVH. Now we need it as an attribute to access in the g-buffer. We're also including a mesh index, which reprojection uses to accept/discard samples. (Beforehand we discarded based on material index, but mesh index is more specific, which means less ghosting between meshes that share the same mats)

src/renderer/glsl/gBuffer.frag
jaxry: `vMaterialMeshIndex` is a float type, but the number its set to is always whole. Some whole numbers can't be exactly represented as a float (e.g. 3 becomes ~2.99999), so taking the floor of `2.99999` would produce `2` when we really want `3`. Thoughts on a better way to handle this? As long as we double check performance, would using an integer attribute be cleaner?

`materialIndex` needs to be an `int` because its use as an index for array lookup.
`meshIndex` can stay as a float since we're passing it into a texture.

src/renderer/glsl/rayTrace.frag
jaxry: This handled portions of the scene where the ray just leaves to infinity without hitting anything. We override the infinite position with a finite one, so we can still use it for reprojection. This would be handy with color-based reprojection in the case of temporal anti-aliasing, but we're not doing this yet so it's just simpler to remove this logic. It will decrease ghosting as well since we're reprojecting less without it.

src/renderer/glsl/rayTrace.frag
jaxry: Yup

src/renderer/glsl/reproject.frag
jaxry: Related to the infinite position code removed above. We don't reproject if the ray escaped to infinity.

src/renderer/glsl/gBuffer.frag
jaxry: I change it to an integer attribute. How does it look? I also made sure to `floor` the index in reproject.frag just to be safe. It didn't seem to make a visual difference though.

src/renderer/glsl/gBuffer.frag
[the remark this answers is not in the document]
lyonsno: As long as we aren't noting performance hits I think this is better.

src/renderer/mergeMeshesToGeometry.js
lyonsno: sounds good.

src/renderer/glsl/rayTrace.frag
lyonsno: gotcha.

PR #66, opened 2020-01-30
Fix glsl error when scene has no materials with maps

lyonsno:
## Brief Description
Scenes with materials without maps give an error when loading since the hybrid refactor, I think due to constructing a vec2 without arguments.

specifically this error:
`Uncaught ERROR: 0:260: 'constructor' : constructor does not have any arguments
WARNING: 0:1206: '1.0e999' : Float overflow`

to replicate, add this code to a test scene:
```
  model.traverse(child => {
    if (child.material) {
      child.material.map = null;
      child.material.normalMap = null;
      child.material.roughnessMap = null;
      child.material.metalnessMap = null;
    }
  });
````

## Pull Request Guidelines


jaxry:
Oops, that would do it.

PR #69, opened 2020-02-06
Fix crash on Windows 10

jaxry:
## Brief Description
Refactors to the shader code in hybrid rendering introduced a driver bug only present on Windows devices. This PR tweaks the shader code so that this crash doesn't happen. The tweak fixes the crashing on my Windows machine, but we should test this on others as well.

## Pull Request Guidelines

### DIFF
--- a/src/renderer/glsl/rayTrace.frag
+++ b/src/renderer/glsl/rayTrace.frag
@@ -90,10 +90,11 @@ source: (defines) => `
     // for (int i = 1; i < defines.bounces + 1, i += 1)
     // equivelant to
     ${unrollLoop('i', 2, defines.BOUNCES + 1, 1, `
-      if (!path.abort) {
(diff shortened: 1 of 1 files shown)


src/renderer/glsl/rayTrace.frag
```diff
@@ -90,10 +90,11 @@ source: (defines) => `
...
     // equivelant to
     ${unrollLoop('i', 2, defines.BOUNCES + 1, 1, `
-      if (!path.abort) {
-        intersectScene(path.ray, si);
-        bounce(path, i, si);
+      if (path.abort) {
+        return vec4(path.li, path.alpha);
       }
+      intersectScene(path.ray, si);
+      bounce(path, i, si);
```
jaxry: ??? Why this works, I don't know ???

PR #71, opened 2020-02-13
Next event estimation

jaxry:
## Brief Description
This PR restructures the path tracing algorithm. In particular, we reduce the number of ray intersections leading to a 2x performance speedup, and a noticeable increase in convergence. This also means the renderer is more responsive, and we're able to render preview buffers at a higher resolution. Most importantly, speeding up the ray tracing pass by 2x lets us spend more time per frame for the spatial denoising filter coming down the line.

**Details**
Our path tracing algorithm was based off the one described in [pbrt](http://www.pbr-book.org/3ed-2018/Light_Transport_I_Surface_Reflection/Path_Tracing.html).

For each bounce:
1. Get direct illumination by importance sampling the light source (the hdr map)
2. Get direct illumination by importance sampling the material brdf.
3. Combine both results using multiple importance sampling (with the power heuristic)
4. Importance sample the brdf to start the next bounce

(1), (2) require shadow ray intersections, and (4) requires a full scene ray intersection. If (4) ends up hitting a light source, we discard the result and terminate the path.

This results in a total of **3 intersections per bounce**

This PR restructures the algorithm to work as follows.

1. Get direct illumination by importance sampling the light source (the hdr map)
2. Importance sample the brdf to start the next bounce
3. If the interesction is
    3.a. *a light source*: Use the power heuristic to add this light's contribution and terminate the path.
    3.b. *a surface*: Start next bounce on this surface.

This results in a total of **2 intersections per bounce**. We're able to do this by effectively combining the (2) and (4) steps in the original algorithm. Combining the steps results in increased performance per frame, but with less information being calculated, resulting in slower convergence per frame. However, the increase in performance outweighs the loss in convergence per frame, leading to a net *increase in convergence* over real time.

This technique is similar to the "next event estimation" technique described in 
* http://www.cs.uu.nl/docs/vakken/magr/2015-2016/slides/lecture%2008%20-%20variance%20reduction.pdf
* https://www.youtube.com/watch?v=bGU_JVyvzLA

These sources don't explicitly mention combining steps (2) and (4), but [these notes](http://www.cs.cornell.edu/courses/cs6630/2012sp/notes/07pathtr-notes.pdf) mention the idea in the section, *Version 1.1*

**Note**
This PR introduces a regression to the shadow catcher. Specifically on non-metallic, specular surfaces, the reflection of the shadow catcher appears somewhat brighter than it should be, with some (but not all) hdr maps. This issue stems from the shadow catcher itself (and not from next event estimation directly), but I've been unable to find a fix for the bug. The issue is subtle enough that it shouldn't be noticeable in most cases.

Master: 10 seconds of rendering - 15 fps
master 15 fps
Next event estimation: 10 seconds of rendering - 30 fps
nee 30 fps

Both renders have stratified sampling *disabled* to increase similarity between each version's random sampling.

The next event estimation version runs at twice the framerate, and has noticeable less noise after running for 10 seconds.

## Pull Request Guidelines


scenes/renderer-test/main.js
```diff
@@ -197,6 +197,8 @@ function init() {
     const geo = new THREE.PlaneBufferGeometry(1000, 1000);
     const mat = new THREE.MeshStandardMaterial();
     mat.shadowCatcher = true;
+    mat.roughness = 0.5;
+    mat.metalness = 0.0;
```
jaxry: Shadow catcher supports roughness and metalness values now! Explicitely set both of these to get intended results. A less rough shadow catcher has more secular reflections of the scene.

src/renderer/glsl/chunks/random.glsl
```diff
@@ -33,4 +33,20 @@ float randomSample() {
...
+
+MaterialSamples getRandomMaterialSamples() {
+  MaterialSamples samples;
+
+  samples.s1 = randomSampleVec2();
+  samples.s2 = randomSampleVec2();
+  samples.s3 = randomSampleVec2();
+
+  return samples;
+}
```
jaxry: Each material sampling function calls this method to get random variables, instead of retrieving them manually. This makes it easier to deal with materials like sampleGlass which don't use all random variables. The stratified samples now stay synced between materials regardless of how many random variables are used.

src/renderer/RenderingPipeline.js
```diff
@@ -30,6 +30,8 @@ export function makeRenderingPipeline({
   // higher number results in faster convergence over time, but with lower quality initial samples
   const strataCount = 6;
 
+  const desiredTimeForPreview = 14;
+
```
jaxry: We can bump up this number from 10 to 14. The 2x increase in performance will already increase the resolution by sqrt(2), but it turns out we're also able to increase this number without losing responsiveness! We should test that things stay responsive in Machete.

scenes/renderer-test/main.js
lyonsno: YAY!!!!
Does it also accept maps?

src/renderer/RenderingPipeline.js
lyonsno: On my machine I can get up to like 30 before experiencing any drop below 60fps! For our use-case!  This is really huge @jaxry, I want to record a video for you.  I've also tweaked a few other params.

src/renderer/glsl/chunks/sampleMaterial.glsl
```diff
@@ -3,123 +3,106 @@
...
-    } else {
-      return li;
-    }
+  vec3 lightDir;
+  vec2 uv;
+  float lightPdf;
+  bool brdfSample = false;
+
+  if (lastBounce && diffuseOrSpecular.x < 0.5) {
+    // reuse this sample by multiplying by 2 to bring sample from [0, 0.5), to [0, 1)
```
lyonsno: Could you elaborate on the advantages of re-using samples like this?

src/renderer/glsl/chunks/sampleMaterial.glsl
```diff
@@ -3,123 +3,106 @@
...
+  cosThetaL = dot(si.normal, lightDir);
 
-  // Get new path direction
+  orientation = dot(si.faceNormal, viewDir) * cosThetaL;
+  path.abort = orientation < 0.0;
 
-  if (lastBounce) {
+  if (path.abort) {
     return;
   }
```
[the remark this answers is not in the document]
jaxry: I think the code was just unorganized before due to repeated additions. But this order seems logical to me too :+1:

src/renderer/glsl/chunks/sampleMaterial.glsl
jaxry: For stratified sampling, you want as few samples as possible to increase the converging effect stratification has. You can reduce samples by transforming 1 sample into 2 independent samples, which helps convergence. The downside is that doing so reduces precision of the two resulting samples. But in this case the samples are binary so the precision loss doesn't have any actual downside.

PR #72, opened 2020-02-18
Distance fog

lyonsno:
Add a fog scale parameter to the module.  Default value is zero, which applies no fog.  Fog works by blending objects into the env map as they appear further away. Values above zero will specify the ray distance at which objects are fully obscured by fog, i.e. replaced by the envmap in that direction.

Also adds a few params which allow for external fine-tuning of temporal filtering and sampling settings .

Screen Shot 2020-02-18 at 9 46 05 AM

## Pull Request Guidelines


src/renderer/RenderingPipeline.js
```diff
@@ -18,18 +18,24 @@ export function makeRenderingPipeline({
...
   // how many partitions of stratified noise should be created
   // higher number results in faster convergence over time, but with lower quality initial samples
   const strataCount = 6;
 
+  const desiredTimeForPreview = renderingParams.previewTime;
+
+  const fogScale = renderingParams.fogScale;
+
+  const bounces = renderingParams.bounces; // number of global illumination bounces
+
```
jaxry: We could destructure these too if you like that idea.

src/renderer/glsl/chunks/intersect.glsl
```diff
@@ -205,6 +205,7 @@ void intersectScene(inout Ray ray, inout SurfaceInteraction si) {
         int materialIndex = floatBitsToInt(r2.w);
         vec3 faceNormal = r2.xyz;
         surfaceInteractionFromBVH(si, tri, hit.barycentric, index, faceNormal, materialIndex);
+        ray.distance = length(ray.o - si.position);
```
jaxry: I believe this value is already stored in `ray.tMax`, so no need to compute it here. It's a weird variable name but it's what pbrt uses.

src/renderer/glsl/chunks/envmap.glsl
```diff
@@ -110,4 +110,17 @@ vec3 sampleBackgroundFromDirection(vec3 d) {
   return textureLinear(backgroundMap, uv).rgb;
 }
 
+vec3 applyFog(vec3 li, float distance, vec3 direction, inout vec3 beta) {
+  #ifdef FOG_SCALE
+
+    float fogWeight = max(0.0, 1.0 - ((max(0.0, distance)) / FOG_SCALE));
+    fogWeight = pow(fogWeight, 2.0);
```
jaxry: If we're always doing a pow of 2.0 because it looks nice, `fogWeight *= fogWeight` would at least be as performant if not more so.

src/renderer/glsl/chunks/envmap.glsl
```diff
@@ -110,4 +110,17 @@ vec3 sampleBackgroundFromDirection(vec3 d) {
   return textureLinear(backgroundMap, uv).rgb;
 }
 
+vec3 applyFog(vec3 li, float distance, vec3 direction, inout vec3 beta) {
+  #ifdef FOG_SCALE
+
+    float fogWeight = max(0.0, 1.0 - ((max(0.0, distance)) / FOG_SCALE));
```
jaxry: I'm not sure how different it looks, but supposedly a more "realistic" fog is exponential fog.

https://catlikecoding.com/unity/tutorials/rendering/part-14/

The weight would be `1 / pow(2.0, distance * scaleConstant)`, where the constant is some low number like 0.001.

jaxry:
Would it make sense to implement [Scene.fog](https://threejs.org/docs/#api/en/scenes/Scene.fog) as best we can? We could use the `near` and `far` values. For `color`, we could either ignore it, or we implement it, and if it's set to `null` that's when we blend with the envmap?

src/RayTracingRenderer.js
```diff
@@ -54,7 +58,15 @@ export function RayTracingRenderer(params = {}) {
 
     const bounces = module.bounces;
 
-    pipeline = makeRenderingPipeline({gl, optionalExtensions, scene, toneMappingParams, bounces});
+    const renderingParams = {
+      reprojectedSamples: module.reprojectedSamples,
+      initialUniformSamples: module.initialUniformSamples,
+      previewTime: module.previewTime,
```
jaxry: I'm hesitant to expose these values to the user

* They are highly dependent on the current pipeline. Any change to the pipeline necessarily requires changes to these numbers, and if the user overrides them, their renderer will break when they update to the latest pipeline.
* They are scene independent. There are ideal numbers that work on the majority of devices that work better than other numbers. It's not opinionated in other words, so if our current numbers are off, it would be better to fix them with a PR than with user-overwritten values.

Thoughts?

lyonsno:
> Would it make sense to implement [Scene.fog](https://threejs.org/docs/#api/en/scenes/Scene.fog) as best we can? We could use the `near` and `far` values. For `color`, we could either ignore it, or we implement it, and if it's set to `null` that's when we blend with the envmap?

yes!  great idea.

src/renderer/glsl/chunks/sampleMaterial.glsl
```diff
@@ -3,123 +3,112 @@
...
+void sampleMaterial(SurfaceInteraction si, int bounce, inout Path path) {
+  bool lastBounce = bounce == BOUNCES;
+  mat3 basis = orthonormalBasis(si.normal);
+  vec3 viewDir = -path.ray.d;
 
-  float lightPdf;
-  vec2 uv;
-  vec3 lightDir = sampleEnvmap(random, uv, lightPdf);
+  float originalRayDistance = path.ray.distance;
+  vec3 originalRayDirection = path.ray.d;
```
jaxry: Instead of copying these values, could we just keep them as they are, and instead use a separate Ray for the shadow intersection?

src/RayTracingRenderer.js
lyonsno: Hmm. I definitely see your point here. It has been very convenient for me to easily modify them in my work when using ray tracing renderer as a dependency though..  Ideas?

jaxry: That makes sense. What about exposing these params prefixed with an underscore (or something to denote they're "private"), and not documenting them in the wiki? That gives us the power to add/remove params when necessary (for example, `previewTime` might not be used with benchmark redesign), and we don't have to worry about supporting them. But we can still tweak the numbers, from a dev perspective, as a dependency to hone in on good values to merge with a PR?

Just throwing ideas here.

PR #73, opened 2020-02-20
Dynamic preview resolution

jaxry:
## Brief Description
This PR changes how the low resolution preview frames scale to a device's performance.

Beforehand, the preview frame's resolution was based on the performance of the tile rendering. This served as only a rough approximation which resulted in the preview resolution being either too high (unresponsive) or too low (poor quality), depending on the scene and hardware.

Now, the performance of each preview frame is measured directly, and adjustments to this resolution are made on the fly. This ideally means that the resolution should close-to-exactly match the performance of the device and scene.

In addition, this PR skips rendering a preview frame upon initialization and goes straight to tile rendering. This should stop the ugly artifacts for appearing on first load.

## Pull Request Guidelines


src/RayTracingRenderer.js
```diff
@@ -136,17 +135,12 @@ export function RayTracingRenderer(params = {}) {
...
-      }
-
+    if(module.maxHardwareUsage) {
+      // render new sample for the entire screen
+      pipeline.drawFull(camera);
     } else {
-      pipeline.drawOffscreenTile(camera);
+      // render new sample for a tiled subset of the screen
+      pipeline.draw(camera);
     }
```
jaxry: Removed some old vestigial code that's no longer supported.

src/renderer/RenderScale.js
```diff
@@ -0,0 +1,70 @@
...
+  const maxRenderbufferSize = gl.getParameter(gl.MAX_RENDERBUFFER_SIZE);
+
+  if (maxRenderbufferSize <= 8192) {
+    return 100000;
+  } else if (maxRenderbufferSize === 16384) {
+    return 200000;
+  } else if (maxRenderbufferSize >= 32768) {
+    return 300000;
+  }
+}
```
jaxry: Similar to the estimates in TileRender, but lower to account for the fact that preview rendering isn't as efficient (due to lack of stratified sampling, tone mapping after every frame, etc).

src/renderer/RenderScale.js
```diff
@@ -0,0 +1,70 @@
...
+    scale.set(renderWidth / fullWidth, renderHeight / fullHeight);
+  }
+
+  let lastTime;
+
+  function onBenchmark(time) {
+    const elapsed = time - lastTime;
+    const pixelsPerTime = renderWidth * renderHeight / elapsed;
+
+    const expAvg = 0.99999;
```
jaxry: Micro-stutters can lead to over-aggressive resizing, so we use a very high moving average to smooth them out. It takes about 1-2 seconds for the preview resolution to adjust to the right number.

src/renderer/RenderingPipeline.js
```diff
@@ -30,7 +31,12 @@ export function makeRenderingPipeline({
   // higher number results in faster convergence over time, but with lower quality initial samples
   const strataCount = 6;
 
-  const desiredTimeForPreview = 14;
+  // tile rendering can cause the GPU to stutter, throwing off future benchmarks for the preview frames
+  // wait to measure performance until this number of frames have been rendered
+  const previewFramesBeforeBenchmark = 5;
```
jaxry: I tested this by moving the mouse in tiny increments. That rapidly switches between tile rendering and preview rendering, and the stutters stopped affecting the benchmark at `3`. I've set this to `5` for some leeway.

jaxry:
I'm getting great results with this change on my hardware, but we should test it on a few more devices to make sure the numbers are tuned correctly

* `timePerFrame` should be set so that the target fps feels smooth without sacrificing too much quality. It's currently set to `21` which is about 48 fps.
* `expAvg` is set so that microstutters don't throw off the benchmark, while also being low enough that the resolution adjusts quickly to changes in the scene. Currently set to `0.99999`
* `previewFramesBeforeBenchmark` is set so that fluctuating between tile rendering and preview rendering (by moving the mouse small amounts) doesn't throw off the benchmark.

src/renderer/RenderSize.js
```diff
@@ -0,0 +1,72 @@
...
+
+  let lastTime;
+
+  function onBenchmark(time) {
+    const elapsed = time - lastTime;
+    const pixelsPerTime = renderWidth * renderHeight / elapsed;
+
+    const expAvg = 0.7;
+    pixelsPerFrame = expAvg * pixelsPerFrame + (1 - expAvg) * timePerFrame * pixelsPerTime;
+    pixelsPerFrame = clamp(pixelsPerFrame, 2048, maxPixelsPerFrame);
```
jaxry: Exponential averaging smooths out microstutters. 0.7 keeps the framerate mostly constant, while still responding quick enough to changes in scene complexity (less than a second).

src/RayTracingRenderer.js
```diff
@@ -133,6 +137,15 @@ export function RayTracingRenderer(params = {}) {
       initScene(scene);
     }
 
+    if (!time) {
+      console.warn('Ray Tracing Renderer warning: render(scene, camera, time) expects a third parameter equal to the time parameter given to the requestAnimationFrame callback.');
```
lyonsno: Maybe add something like "to avoid poor performance please supply this parameter"

jaxry: Yeah, something like that would be good. The PR still needs a few adjustments and once I make those I'll make comments here with my questions and stuff.

src/RayTracingRenderer.js
```diff
@@ -37,7 +37,6 @@ export function RayTracingRenderer(params = {}) {
     needsUpdate: true,
     onSampleRendered: null,
     renderWhenOffFocus: true,
-    renderToScreen: true,
```
lyonsno: What's the history here?

src/renderer/RenderSize.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  function adjustSize(elapsedMs) {
+    if (!elapsedMs) {
+      return;
+    }
+
+    const error = desiredMsPerFrame - elapsedMs;
+
+    pixelsPerFrame += 600 * error;
```
lyonsno: what does the number 600 represent?  Are you tweaking it manually? maybe assigning a variable name could help with readability.

src/RayTracingRenderer.js
jaxry: This was back when we explored the idea of rendering offscreen, and displaying the finished, high-sample render once it was ready. Probably not that useful now that we're moving towards full real-time?

src/renderer/RenderSize.js
jaxry: It's an arbitrary number that controls the strength of changes to the pixel count, tuned so that it doesn't fluctuate when the camera stays in the same area, but adjusts fast enough when the scene complexity changes. Agreed, a variable name is better.

jaxry:
@lyonsno what are your thoughts about how the API should look like? What would be the least intrusive way to introduce this change?

My thoughts:
1. `renderer.render(scene, camera, time)`
Pros: Only requires one extra argument for an otherwise unmodified API. 
Cons: Conflicts with WebGLRenderer's function signature. WebGLRenderer accepts a third (depricated) parameter for some unrelated task, and supplying it a `time` parameter will break something.

2. `renderer.time(time); renderer.render(scene, camera);`
Pros: Leaves `render()` in an unmodified state that's compatible with both classes.
Cons: Introduces a new, mandatory method that needs to be called before `render()`. That adds more cognitive load IMO.

Neither of these solutions seem ideal to me, but I'm leaning towards 2. a little more. 

Do you think it's best to avoid a mandatory time variable altogether? We can achieve an accurate FPS if we divide a few seconds of time by the number of frames rendered during that time. It's something we could do internally. But that gives us delayed performance updates, and it sounds really hard to program to benchmark preview windows and tiles separately if it takes a few seconds for the benchmark to record for each case. But if this new API is too much of a headache, it could be worth looking into.

src/RayTracingRenderer.js
```diff
@@ -138,13 +135,14 @@ export function RayTracingRenderer(params = {}) {
     }
 
     if (!time) {
```
elfrank: can `time` be `0`?

src/RayTracingRenderer.js
```diff
@@ -111,13 +104,20 @@ export function RayTracingRenderer(params = {}) {
...
-    }
+  let validTime = 1;
+
+  function restartTimer() {
+    validTime = NaN;
+  }
+
+  let time;
+
+  module.time = (t) => {
```
elfrank: My two cents: use a separate `renderer.time` method vs adding a 3rd parameter in `renderer.render` (what you are doing already).

I think `renderer.render` should follow the same signature as `three.js`
`renderer.time` might evolve in the future (we don't know enough yet).

**Suggestion 1**: consider using a verb for the function name, such as `synchronize`
```javascript
module.synchronize = (t) => {
  time =  t;
};
```

**Suggestion 2**: (nit) Updating variable names might provide additional clarity, such as `currentTime` and `lastTime`.

jaxry:
That's awesome to hear! Thanks for testing so extensively. I'll test that a preview target of 18ms works well on my devices too.

jaxry:
Alright, let's go with these numbers then. It'll be quick to make another PR if it needs any tweaking.

PR #77, opened 2020-03-16
Performance tiers

lyonsno:
## Brief Description
This is a draft PR to share a temp branch.
The idea is to provide functionality to force the renderer to set it's performance as if it were running on devices with different hardware levels.
Numbers are not finalized.
I'm also planning to add debug output so we can see what the dynamic algorithm comes up with for various devices.

TODO or DELETE

## Pull Request Guidelines

### DIFF
--- a/build/RayTracingRenderer.js
+++ b/build/RayTracingRenderer.js
@@ -7,21 +7,21 @@
   const ThinMaterial = 1;
   const ThickMaterial = 2;
   const ShadowCatcherMaterial = 3;
-  const MinimumPerformance = 4;
-  const OkPerformance = 5;
-  const GoodPerformance = 6;
-  const ExcellentPerformance = 7;
-  const DynamicPerformance = 8;
+  const MinimumRayTracingPerformance = 4;
+  const OkRayTracingPerformance = 5;
+  const GoodRayTracingPerformance = 6;
+  const ExcellentRayTracingPerformance = 7;
+  const DynamicRayTracingPerformance = 8;
 
   var constants = /*#__PURE__*/Object.freeze({
     ThinMaterial: ThinMaterial,
     ThickMaterial: ThickMaterial,
     ShadowCatcherMaterial: ShadowCatcherMaterial,
-    MinimumPerformance: MinimumPerformance,
-    OkPerformance: OkPerformance,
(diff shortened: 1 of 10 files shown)


src/constants.js
```diff
@@ -1,3 +1,8 @@
 export const ThinMaterial = 1;
 export const ThickMaterial = 2;
 export const ShadowCatcherMaterial = 3;
+export const MinimumRayTracingPerformance = 4;
```
elfrank: Suggestion:

create a namespace called `PerformanceTier` or similar, an place values there. For example

```javascript
// file name: PerformanceTier
export const UNKNOWN = X; // ??
export const BAD = XX;
export const GOOD = Y;
export const GREAT = YY;
export const EXCELLENT = Z;
export const DYNAMIC = ZZ;
```

that way, you would access the variables as:
```javascript
PerformanceTier.UNKNOWN
PerformanceTier.BAD
PerformanceTier.GOOD
PerformanceTier.GREAT
PerformanceTier.EXCELLENT
PerformanceTier.DYNAMIC
```

src/renderer/RenderSize.js
```diff
@@ -54,6 +57,21 @@ export function makeRenderSize(gl) {
   };
 }
 
+function pixelsPerFrameFromPerformanceLevel(performanceLevel) {
```
elfrank: `nit`: add docs for method

build/RayTracingRenderer.js
```diff
@@ -4083,6 +4116,21 @@ void sampleGlassSpecular(SurfaceInteraction si, int bounce, inout Path path) {
     };
   }
 
+  function pixelsPerMsFromPerformanceLevel(performanceLevel) {
```
elfrank: how can we ensure consistency between `pixelsPerFrameFromPerformanceLevel` and `pixelsPerMsFromPerformanceLevel`? do we need two methods?

jaxry:
Am I understanding correctly that there's no intent to merge this PR? If so, I don't think we need to worry about code structure as long as it gets the job done for the purposes of testing.

jaxry:
Closing for now, since I think we don't need this anymore?

PR #81, opened 2020-04-06
Fix recursive loop in BVH building

ToJans:
## Brief Description
When all primitive centroids are equal during a recursive BVH building step, it results in a recursive loop.
This fix detects this case and returns a leaf node instead.

## Pull Request Guidelines

### DIFF
--- a/src/renderer/bvhAccel.js
+++ b/src/renderer/bvhAccel.js
@@ -149,6 +149,10 @@ function recursiveBuild(primitiveInfo, start, end) {
     }
     const dim = maximumExtent(centroidBounds);
 
+    if (centroidBounds.max[dim] === centroidBounds.min[dim]) {
+      return makeLeafNode(primitiveInfo.slice(start, end), bounds);
+    }
+
     let mid = Math.floor((start + end) / 2);
 
     // middle split method


jaxry:
Thanks for your PR! I encountered this bug myself, but I never got to the bottom of it, so it's great you figured it out.

I have a lot on my plate right now, but I can review and merge your fix in the next couple days when I have a moment.

src/renderer/bvhAccel.js
```diff
@@ -215,6 +215,12 @@ function recursiveBuild(primitiveInfo, start, end) {
         }
         return b <= minCostSplitBucket;
       }, start, end);
+      
+      // Avoid a recursive loop
+      // This happens when all primitive centroids are equal
+      if ( mid === end) {
+        return makeLeafNode(primitiveInfo.slice(start, end), bounds);
+      }
```
jaxry, line 223: Is it also possible that `mid === start`, and would that lead to an infinite loop as well?

I reviewed the chapter in [pbrt](http://www.pbr-book.org/3ed-2018/Primitives_and_Intersection_Acceleration/Bounding_Volume_Hierarchies.html) and there's a section dealing with centroids in the same position (ctrl+f "If all of the centroid points are at the same position). 

Their solution would be to add the logic at line https://github.com/hoverinc/ray-tracing-renderer/blob/2b2ab926743f48e8c8185f91c7cea826a092aa4a/src/renderer/bvhAccel.js#L153

```javascript
if (centroidBounds.max[dim] === centroidBounds.min[dim]) {
   return makeLeafNode(primitiveInfo.slice(start, end), bounds);
}
```
If that works, implementing this logic would save us from computing the SAH just for it to fail, and it would deal with the `start === mid` case as well.

I'm not able to test it though. I couldn't find a model with this issue. Could I ask what model you were using that exposed the bug?

ToJans: Your proposal would indeed be better. 

I am not well versed in the pbrt book, but I was trying to incorporate this engine in one of my own products, and noticed some models generated infinite loops, while others worked just fine.
If you want to, you can play with a demo (using webgl) yourself [here](https://productconfigurator.virtualsaleslab.com/r/demo/vsl/en/designer/index/reynaersconservatory).
PS: you can export a 3D model by pressing CTRL-ALT-O, but just do not tell anyone, it is supposed to be a debugging feature, so it might stop working in the future...

jaxry: Apologies for the late response. If that pbrt snippet I provided works with the problematic models, would you consider committing it in this PR instead of what we already have? Or if it doesn't work, let's stick with your solution. Either way, let me know and we can get this merged.

ToJans: It took a while to change this, but I made the change as requested. It worked perfectly with my model.

jaxry:
Looks good! Lastly, we omit the build files from PRs since they're automatically generated for each release. Could you revert the changes to the build files?

ToJans:
I removed the `/build` folder changes from the commit. Let us hope we are good to go now!

elfrank:
@ToJans thanks!

PR #83, opened 2020-04-20
Light separation / Albedo demodulation

jaxry:
## Brief Description
This PR separates light output from the ray tracing stage into separate diffuse and specular light buffers. In addition, albedo is demodulated from the ray tracing step and is added during the post-processing stage. This technique is a necessary precursor for spatial filtering, and it also improves visual quality across the board.

Note: This PR reduces performance by ~10% on mid-range devices. However, the advantages the technique brings outweigh the performance penalty.

Advantages:
* Improved temporal filter. As a result of the light separation, this PR now temporally filters different amounts for diffuse and specular light. This means diffuse surfaces, which reflect light the same amounts in all directions, can use a long temporal filter. And specular surfaces can filter in line with their roughness property. This drastically reduces noise and improves visual quality.

* Full resolution textures. Albedo is now always rasterized at full resolution, since its now only applied in the post-processing step.

This PR also opens the door for some fast to implement future improvements
* Mip mapping. Since albedo is now entirely rasterized, it supports mip mapping and anisotropic filtering. This can be enabled to get rid of aliasing from textures. But since we still use progressive rendering, the max mip level needs to be increased as frames are rendered so as not to overblur textures.

* Improve temporal reprojection. Currently, history samples are only rejected if the mesh IDs don't match. This leads to a lot of temporal artifacts for models with shared geometry in a single mesh, which is made especially prevalent by the increased temporal filtering strength of this PR. We can improve the reprojection by rejecting history if positions and normals don't match.

Lastly, the shadow catcher is no longer supported. It is tricky (but not impossible) to implement with decoupled lighting. We are thinking it might make the most sense to remove the shadow catcher from the renderer and instead add support for [Three.js fog](https://threejs.org/docs/#api/en/scenes/Fog). Fog can achieve similar looks to the shadow catcher, but in a more physically based way.

## Pull Request Guidelines


src/renderer/AlbedoSeparationPass.js
```diff
@@ -0,0 +1,50 @@
...
+
+    renderPass.useProgram();
+    fullscreenQuad.draw();
+  }
+
+  return {
+    draw,
+    outputLocs: renderPass.outputLocs
+  };
+}
```
jaxry, line 50: Since we decoupled albedo from ray tracing, it needs to be antialiased in its own pass.

src/renderer/Framebuffer.js
```diff
@@ -13,34 +13,52 @@ export function makeFramebuffer(gl, { color, depth }) {
...
+  const glLocation = gl.COLOR_ATTACHMENT0 + location;
+
+  if (layer !== undefined) {
+    gl.framebufferTextureLayer(gl.FRAMEBUFFER, glLocation, texture.texture, 0, layer);
+  } else {
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, glLocation, texture.target, texture.texture, 0);
+  }
+
+  return glLocation;
+}
```
jaxry, line 64: * Refactor the Framebuffer API to take an explicit `location` property, instead of an implicit location given as the key to an object.

* Add support for array texture targets

src/renderer/MaterialBuffer.js
```diff
@@ -125,25 +126,25 @@ function maxImageSize(images) {
...
-  materialBuffer.set('Materials.diffuseNormalRoughnessMetalnessMapIndex[0]', interleave(
+  materialBuffer.set('Materials.diffuse_normal_roughness_metalness_mapIndex[0]', interleave(
     { data: bufferData.diffuseMapIndex, channels: 1 },
     { data: bufferData.normalMapIndex, channels: 1 },
     { data: bufferData.roughnessMapIndex, channels: 1 },
     { data: bufferData.metalnessMapIndex, channels: 1 }
   ));
 
-  materialBuffer.set('Materials.diffuseNormalMapSize[0]', interleave(
+  materialBuffer.set('Materials.diffuse_normal_mapSize[0]', interleave(
```
jaxry, line 147: The material buffer packs  several different variables into vec4. The naming made this confusing, so I added underscores between each variable in the vec4. Does this help?

src/renderer/RenderingPipeline.js
```diff
@@ -79,13 +87,14 @@ export function makeRenderingPipeline({
...
-  let screenWidth = 0;
-  let screenHeight = 0;
+  const screenSize = new Vector2();
 
   const fullscreenScale = new Vector2(1, 1);
 
   let lastToneMappedScale = fullscreenScale;
 
+  // TODO: Move frame buffer creation to respective program file
+
```
jaxry, line 97: Framebuffer initialization requires a bunch of module-level variables, and it's getting really messy. I believe moving framebuffers into their respective shader passes would clean it up and aid in the separation of concerns. I'd like to make this change as a separate PR.

src/renderer/Texture.js
```diff
@@ -80,15 +97,19 @@ export function makeTexture(gl, params) {
...
+    gl.texStorage2D(target, levels, internalFormat, width, height);
     if (data) {
       gl.texSubImage2D(target, 0, 0, 0, width, height, format, type, data);
     }
   }
 
+  if (useMipmaps) {
+    gl.generateMipmap(target);
+  }
+
```
jaxry, line 112: Add support for mip mapping. Not currently used by the renderer, but it will be.

src/renderer/glsl/chunks/toneMapOperators.glsl
```diff
@@ -0,0 +1,32 @@
...
+    // optimized filmic operator by Jim Hejl and Richard Burgess-Dawson
+    color = max(vec3( 0.0 ), color - 0.004);
+    return pow((color * (6.2 * color + 0.5)) / (color * (6.2 * color + 1.7) + 0.06), vec3(2.2));
+  }
+
+  // https://knarkowicz.wordpress.com/2016/01/06/aces-filmic-tone-mapping-curve/
+  vec3 acesFilmic( vec3 color ) {
+    return clamp((color * (2.51 * color + 0.03)) / (color * (2.43 * color + 0.59) + 0.14), vec3(0.0), vec3(1.0));
+  }
+`;
```
jaxry, line 32: Move tone mapping operators to their own file

scenes/sample-models/main.js
```diff
@@ -108,7 +108,7 @@ function createGroundMesh() {
   mat.color.set(0xffffff);
   mat.roughness = 0.5;
   mat.metalness = 0.0;
-  mat.shadowCatcher = true;
+  // mat.shadowCatcher = true;
```
jaxry, line 111: In place until I get the fog feature working.

carlos-lopez-garces:
I ran all the example scenes on my machine:

```
Windows 10
Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz, 2601 Mhz, 6 Core(s), 12 Logical Processor(s)
GeForce RTX 2060
```

My observations are not based on precise measurements:

- The render converges very fast, faster than I remember.
- The Lee Perry Smith Head and Georgia Tech Dragon look very smooth on the lit areas, and edges are super sharp. Very little grain in the shadowed areas.
- Speed of convergence was the most evident in the PBR spheres scene. Less than 3 seconds.
- I don't remember the renderer-test scene looking as great as it does now.
- No errors on the console.

Like you said, there's no more shadow catcher. I believe that Design Pro scenes use it. Would we be able to merge this PR then?

renderer-test
sample-models_cornell-box
sample-models_georgia-dragon_sunset
sample-models_lee-perry_sunset
sample-models_pbr-spheres
sample-models_pbr-spheres_sunset
webgl-comparison

jaxry:
@carlos-lopez-garces Thanks a lot for giving this a run. It's good to confirm it runs well on high end devices.

I've been really confused about the next possible steps for the renderer. Everything in this PR will be used in the future, so I think it's great to review and to confirm everything works.. But I need a couple days to do some more investigation into spatial denoising. After that I can summarize my thoughts to the team and we can figure out if we want to merge this PR before or after spatial denoising.

>Like you said, there's no more shadow catcher. I believe that Design Pro scenes use it. What would be the alternative for Design Pro?
Noah was prototyping it, and we think a simple fog implementation that blends the render with the environment map could replace the shadow catcher.  It shouldn't be too hard to implement proper.

Issue #92, opened 2020-05-08
Reflection problem


VinAr22:
Hello! I updated the ray tracing version from 0.9.0 to 0.9.1 and got some reflection problems on my model. It seems that reflections are not projected onto the model properly and have an offset on the model.

Please, see below

v0.9.1 
![0 9 1 right](https://user-images.githubusercontent.com/49755663/81381421-e8e7fa80-9136-11ea-98ce-f014dc1ed1a6.jpg)

v0.9.0
![0 9 0 right](https://user-images.githubusercontent.com/49755663/81381432-ebe2eb00-9136-11ea-93f9-2e13ee72aca6.jpg)

v0.9.1 
![0 9 1 left shadow](https://user-images.githubusercontent.com/49755663/81381472-fac99d80-9136-11ea-90d1-a2ab815cb09e.jpg)
![0 9 1 left](https://user-images.githubusercontent.com/49755663/81381604-2ea4c300-9137-11ea-99c3-52ed125e2c61.jpg)

v0.9.0
![0 9 0 left shadow](https://user-images.githubusercontent.com/49755663/81381487-01f0ab80-9137-11ea-8c9b-afe6f7ce515c.jpg)
![2](https://user-images.githubusercontent.com/49755663/81382057-f2be2d80-9137-11ea-859e-163adce3724d.jpg)

Top projection. Its just the plane 1000x1000, but reflects like rectangle
![0 9 1 top](https://user-images.githubusercontent.com/49755663/81381504-087f2300-9137-11ea-9784-6d01388c59c2.jpg)

p.s.
do not pay attention to different exposures, I change it manually

jaxry:
Thanks for submitting this issue. It looks like it might be the same as https://github.com/hoverinc/ray-tracing-renderer/issues/90. Weirdly enough, the test scene has several specular materials, but the bug isn't present there.

I think https://github.com/hoverinc/ray-tracing-renderer/commit/0324fefbb9fb018948a8ed9e7afc42e6e3620eb7 could somehow be the culprit. Would you be able to share that scene as a GLTF (using the [GLTFExporter](https://threejs.org/docs/#examples/en/exporters/GLTFExporter) perhaps)? I could try debugging it, unless you have some clues as to what's going on?

VinAr22:
Yes, I will try to create a sample scene tomorrow. Thanks for the response!

jaxry:
[...]

PR #93, opened 2020-05-08
Fix incorrect BVH

jaxry:
## Brief Description
Fixes incorrect BVH generation as experienced in https://github.com/hoverinc/ray-tracing-renderer/issues/92.

There are certain cases where the centroid bounds can be zero but the BVH can still be partitioned. This special case is not present in the test scene, but it is preset in the readme scene with the flight mask. This PR makes the bounds check to happen only if trying to partition using the SAH and not nthElement.

This change definitely fixes the incorrect BVH, but it may re-introduce infinite loops. I have no way of testing this since I don't have access to a problematic model, and I'm not sure how to generate one. Would you have any ideas @ToJans ?

Before: 
Screenshot from 2020-05-08 12-54-07

After: 
Screenshot from 2020-05-08 12-53-18

## Pull Request Guidelines

    - [x] I have added tests to cover my changes, of which pass.

### DIFF
--- a/scenes/renderer-test/main.js
+++ b/scenes/renderer-test/main.js
@@ -196,6 +196,32 @@ function init() {
     model.add(mesh);
   }
 
+  let unreadyMat;
+  {
+    // Create a test (non-buffer) Geometry
+    const geo = new THREE.BoxGeometry(20, 6, 6);
+    const mat = new THREE.MeshStandardMaterial();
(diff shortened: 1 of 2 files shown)


ToJans:
@jaxry Sorry that I was not able to verify this. I can currently not export to GLTF because I am in the process of updating from my old THREE.js version, which does not have a GLTF exporter yet.

PR #96, opened 2020-05-11
Reprojection for spatial denoising

jaxry:
## Brief Description
This PR modifies the temporal reprojection algorithm to closely resemble SVGF. Namely, instead of rejecting sample history if mesh IDs don't match, we instead reject sample history if the depth buffer or normal buffer don't match.

## Pull Request Guidelines

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,3 +1,4 @@
-export const ThinMaterial = 1;
-export const ThickMaterial = 2;
-export const ShadowCatcherMaterial = 3;
+export const StandardMaterial = 1;
+export const ThinMaterial = 2;
+export const ThickMaterial = 3;
+export const ShadowCatcherMaterial = 4;
--- a/src/renderer/RenderingPipeline.js
+++ b/src/renderer/RenderingPipeline.js
@@ -126,7 +126,7 @@ export function makeRenderingPipeline({
     reprojectBuffer = makeHdrBuffer(reprojectPass.outputLocs);
     reprojectBackBuffer = makeHdrBuffer(reprojectPass.outputLocs);
 
-    const normalBuffer = makeTexture(gl, { width, height, storage: 'halfFloat' });
+    // const normalBuffer = makeTexture(gl, { width, height, storage: 'halfFloat' });
     const faceNormalBuffer = makeTexture(gl, { width, height, storage: 'halfFloat' });
     const albedoBuffer = makeTexture(gl, { width, height, storage: 'byte', channels: 4});
     const matProps = makeTexture(gl, { width, height, storage: 'byte', channels: 2 });
@@ -135,6 +135,7 @@ export function makeRenderingPipeline({
 
     function makeGBuffer() {
       const positionBuffer = makeTexture(gl, { width, height, storage: 'float' });
+      const normalBuffer = makeTexture(gl, { width, height, storage: 'halfFloat' });
 
       return makeFramebuffer(gl, {
         colorAttachments: [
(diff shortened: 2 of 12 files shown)


src/constants.js
```diff
@@ -1,3 +1,4 @@
-export const ThinMaterial = 1;
-export const ThickMaterial = 2;
-export const ShadowCatcherMaterial = 3;
+export const StandardMaterial = 1;
```
jaxry, line 1: Start material ids at 1 instead of 0. Now 0 designates that the material is invalid.

src/renderer/RenderingPipeline.js
```diff
@@ -37,7 +37,7 @@ export function makeRenderingPipeline({
 
   // tile rendering can cause the GPU to stutter, throwing off future benchmarks for the preview frames
   // wait to measure performance until this number of frames have been rendered
-  const previewFramesBeforeBenchmark = 2;
+  const previewFramesBeforeBenchmark = 3;
```
jaxry, line 40: The extra work performed by light separation makes the benchmarking more finicky. Once spatial denoising is in, I expect we'll do benchmarking differently so this isn't a final change.

src/renderer/RenderingPipeline.js
```diff
@@ -399,7 +403,7 @@ export function makeRenderingPipeline({
           lightScale: fullscreenScale,
           previousLight: reprojectBackBuffer.color[0],
           previousLightScale: previewSize.scale,
-          reprojectPosition: sampleCount === 0 // Only blend frames and don't reproject positions after camera stays still
+          reprojectPosition: sampleCount <= 1 // Only blend frames and don't reproject positions after camera stays still
```
jaxry, line 406: Was a bug. Sample count will always be at least 1 here.

src/renderer/RenderingPipeline.js
```diff
@@ -441,17 +445,20 @@ export function makeRenderingPipeline({
...
+      useJitter = false;
     } else {
       sampleCount++;
+      useJitter = true;
     }
 
     setCameras(camera, lastCamera);
 
-    updateSeed(screenSize, true);
+    updateSeed(screenSize, useJitter);
```
jaxry, line 461: Conditionally deactivate jittering for fullscreen debug rendering, like we're doing with tile rendering.

src/renderer/glsl/chunks/rayTraceCore.glsl
```diff
@@ -1,8 +1,8 @@
 export default `
-  #define STANDARD 0
-  #define THIN_GLASS 1
-  #define THICK_GLASS 2
-  #define SHADOW_CATCHER 3
+  #define STANDARD 1
+  #define THIN_GLASS 2
+  #define THICK_GLASS 3
+  #define SHADOW_CATCHER 4
```
jaxry, line 5: These same constants are duplicated in JS. That's a code smell and we want to rely on a single source of truth instead. Future refactor.

src/renderer/glsl/chunks/surfaceInteractionDirect.glsl
```diff
@@ -25,7 +19,8 @@ export default `
     vec4 matProps = texture(gMatProps, coord);
     si.roughness = matProps.x;
     si.metalness = matProps.y;
+    si.materialType = int(matProps.z);
 
-    si.hit = meshIndex > 0.0 ? true : false;
+    si.hit = si.materialType > 0 ? true : false;
```
jaxry, line 24: We're no longer using mesh ids, so now we determine whether or not to ray trace a pixel by whether the material type is defined (greater than 0)

src/renderer/glsl/gBuffer.frag
```diff
@@ -47,11 +48,15 @@ source: `
       normal = getMatNormal(materialIndex, uv, normal, dp1, dp2, duv1, duv2);
     #endif
 
-    out_position = vec4(vPosition, float(meshIndex) + EPS);
-    out_normal = vec4(normal, materialType);
-    out_faceNormal = vec4(faceNormal, 0);
+    float linearDepth = vClipPos.z;
+    float linearDiff = max(abs(dFdx(linearDepth)), abs(dFdy(linearDepth)));
```
jaxry, line 52: The derivative of depth across screen space. This scales the rejection sensitivity. e.g. geometry with naturally high depth fluctuations are more lenient to differences in depth.

src/renderer/glsl/gBuffer.frag
```diff
@@ -47,11 +48,15 @@ source: `
...
+    float normalWidth = length(fwidth(normal));
```
jaxry, line 53: Same with normals.

